PR #318, opened 2015-12-03
Fixed an off by one error that would have excluded a null byte.

block8437:
Also for some reason two spaces got stripped off accidentally (Atom does it automatically I think).

### DIFF
--- a/src/crt/string.c
+++ b/src/crt/string.c
@@ -22,8 +22,8 @@
 
 char* strdup(const char* string)
 {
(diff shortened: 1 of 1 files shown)


block8437:
Fixed it to include the null byte in the copy and it will also account for the size of a character in the malloc/copy.

alfred-bratterud:
Hi, great, thanks alot! Could you please reissue the pul-request to the dev-branch? That way we'll have Jenkins do an automatic install / build /  test from scratch.

PR #320, opened 2015-12-03
strdup OBO Fix + Documentation Updates

block8437:
The off by one fix for the strdup function and the documentation updates so tagged along. If the documentation changes are a problem I can fix.


alfred-bratterud:
Great, thanks!

PR #323, opened 2015-12-04
Augmented file...

RicoAntonioFelix:
### File: ` api/utility/membitmap.hpp`
* Changed constructor body to default keyword
* Initialized members in-class
* Changed type of index from signed to unsigned
* Changed C-Style cast to C++-Style cast


RicoAntonioFelix:
Transferred PR to `hioa-cs:dev`

PR #324, opened 2015-12-04
Augmented file...

RicoAntonioFelix:
### File: ` api/utility/membitmap.hpp`
* Changed type of index from signed to unsigned
* Changed constructor body to default keyword
* Changed C-Style cast to C++-Style cast
* Initialized members in-class

### DIFF
--- a/api/utility/membitmap.hpp
+++ b/api/utility/membitmap.hpp
@@ -35,7 +35,7 @@ namespace fs
     typedef uint32_t  index_t;
     static const int CHUNK_SIZE = sizeof(word) * 8;
     
-    MemBitmap() {}
+    MemBitmap() = default;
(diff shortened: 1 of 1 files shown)


alfred-bratterud:
Hi, thanks for contributing. Most of this looks good- but I don't think we want to remove the unsigned from our index_t. @fwsGonzo - please comment.

fwsGonzo:
There is a suspicious return -1 there, so either it must throw some error (which we especially don't want an inline find_first_zero_bit inline function to do). So we have to either change it to signed, or simply make the function return a new free index. The membitmap is supposed to be static, used with the upcoming virtual filesystem, so.. my vote is also for signed.  :+1:

PR #330, opened 2015-12-07
Added new project template

kybeg:
The project template mimics the demo service, but is ready for new projects. The extra script new.sh
itializes the former two files and creates a small 'hello world' example. The README has the documentation.

### DIFF
--- a/examples/project_template/Makefile
+++ b/examples/project_template/Makefile
@@ -1,0 +1,134 @@
+#################################################
+#          IncludeOS SERVICE makefile           #
+#################################################
+
+# The name of your service
+SERVICE = TEMPLATE
+
+# Your service parts
+FILES = START_FILE
+
+# IncludeOS location
+ifndef INCLUDEOS_INSTALL
+	INCLUDEOS_INSTALL=$(HOME)/IncludeOS_install
+endif
+
+# Shorter name
+INSTALL = $(INCLUDEOS_INSTALL)
+
+# Compiler/Linker
+###################################################
+
+OPTIONS = -Ofast -msse3 -Wall -Wextra -mstackrealign 
+
+# External Libraries
+###################################################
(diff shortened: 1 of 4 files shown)


alfred-bratterud:
Hi, thanks! But, a couple of thoughts:

This seems to overlap with the "./seed" folder, its intended use case, as described on the front README.  Since the word 'template' should probably be avoided due to its meaning in C++, I'd suggest keeping the script, just basing it on `./seed` instead. 

The script can live in ./etc for now, but it would also be really useful if we could use it from anywhere. We'd definitely want peoples IncludeOS projects to live outside of our repo, so that they can become repos of their own. My `sed` skills are poor - will this work with any target path? In that case you could even install it to /usr/local/bin if you wanted and use it to create new projects anywhere. If it's not that general, I'd just limit the scope to be like:
`$ includeos_new ~/fancy_web_service`

Note also that there is an ongoing effort around a package manager, which @AndreasAakesson is doing. That wil take over this functionality eventually, but it's nice to have something before that.

kybeg:
Hello, I can do the adjustments. Just drop the pull request.

Just to be clear on the main ideas: 

- A new script called includeos_new in the etc/ folder

- The script will take a project name as argument and create a new folder that contains:
	+ A Makefile
	+ A basic cpp file (is there a strong wish to keep it named service.cpp? I thought maybe using the name of the project as a filename. Regardless, the image will be named the same as the project.
	+ A run.sh script for starting the VM image (this will still be platform dependent for now)

- The three files will be copied from the seed directory and slightly modified based on the name of the project. 

Correct?

alfred-bratterud:
Yes, sounds great! The name service.cpp isn't important as long as the makefile knows about it and compiles it. You can just commit directly to the same fork so we keep this thread, or close the pull request if you prefer.

PR #332, opened 2015-12-08
Improved OS X script that builds successfully with the correct prerequisites

AndreasAakesson:
Script only looking in `/usr/local/include/` for elf.h (required to make vmbuild) and fetch it to the same location if missing.

Deleted `/src/elf/elf.h` as told [here](https://github.com/hioa-cs/IncludeOS/pull/326#issuecomment-162699046).


alfred-bratterud:
Hi, good - you can actually remove the whole elf folder as well. We could consider just adding that elf.h to our own repo. It's lightly modified for mac, but I think it has everything we need. It's GPL so we can distribute it freely. @fwsGonzo / @kybeg you agree? If so Andreas, just put it in the vmbuilder folder. This communicates clearly that we only use it there, which means the Apache license goes for everything pertaining to IncludeOS services.

fwsGonzo:
I agree

AndreasAakesson:
I'll make a new pull request with these changes.

AndreasAakesson:
.... or I'll just update the title on this one ^_^

AndreasAakesson:
IncludeOS and demo_services now fully build and compile on my machine (OS X 10.9.5 Mavericks).

Prerequisites:
* [homebrew](https://brew.sh)
* `/usr/local` directory with write access.
* `/usr/local/bin` added to PATH.

To compile services/run test:
```
export LD_INC=ld-i686
./test.sh
```
Will fail at qemu. See `etc/vboxrun.sh` for running the service (successful try on my machine).

alfred-bratterud:
Great, we'll see what Jenkins says :-)

kybeg:
I can confirm that it works on OS X El Capitan with Xcode 7.1.1 :)

alfred-bratterud:
Great, thanks!

PR #334, opened 2015-12-11
Refactored code to use modern C++ style

RicoAntonioFelix:
* Changed syntax usage to differentiate initialization from assignment
* Replaced usage of `std::endl` for `\n` to minimize flushes
* Avoid using a raw pointer for disk image buffer by using a `std::vector`
* Used `std::fstream` type objects to avoid manual closing of file handles
* Used C++-Style cast instead of C-Style casts

### DIFF
--- a/vmbuild/vmbuild.cpp
+++ b/vmbuild/vmbuild.cpp
@@ -15,222 +15,216 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <vector>
+#include <fstream>
 #include <iostream>
-#include <sys/stat.h>
+
+#include <elf.h>
 #include <errno.h>
 #include <stdio.h>
-#include <elf.h>
 #include <stdlib.h>
+#include <sys/stat.h>
 
 #define SECT_SIZE 512
-
+#define SECT_SIZE_ERR 99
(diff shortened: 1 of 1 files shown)


alfred-bratterud:
Hi, looks good. One thing: In general I didn't think `endl` should be outlawed - and I just now realize that you're the guy who wrote a lot of the stuff I read in this [good, intelligent discussion about it](https://github.com/isocpp/CppCoreGuidelines/issues/357) :-) Here in the VMbuilder I don't think it's a problem to use it. Inside a for-loop, sure. But, when I see the Jedi Master himself weighing in, and making it gospel, we should follow our own policy of following the guidlines. So +1 to you. 

We do try to use `printf`, or macros derived from it (i.e. INFO / debug), consistently throughout the kernel, but that's because streams are a bit big and we want the OS library to be as small as possible. (we should look into replacing those macros with some fancy template magic - but it has to ensure that we can positively eradicate them from code before production)

Did you test this implementation? If so I'll merge and let Jenkins verify.

RicoAntonioFelix:
I've managed to get the following output in the attached screen-shot...
This is the only result I can use at the moment to say that it works...

Screenshot

alfred-bratterud:
Looks good!

PR #338, opened 2015-12-17
Use Vagrant to standardize build environment

matiashf:
These commits also fix two minor snags I encountered while setting up IncludeOS for the first time.


alfred-bratterud:
Great, thanks!

PR #343, opened 2016-01-01
remove vmbuild's dependency on system headers so it builds on OS X

marssaxman:
vmbuild currently includes elf.h (from a system path) which does not exist on OS X. There was a copy of elf.h in src/elf which does not appear to be used anywhere, so I moved it into vmbuild/ and made a couple of small edits to remove dependencies on other headers.


marssaxman:
I didn't mean for the second commit to become part of the same pull request, but here we are. The second commit makes etc/install_osx.sh more resilient against variations in initial machine state: it creates the IncludeOS_install directory if it doesn't already exist, installs nasm, and specifically identifies the versions of clang installed by brew (version 3.6). In addition, it exports CC_INC and CPP_INC so they will be picked up by src/Makefile.

alfred-bratterud:
Hi, thanks a lot for contributing!  We have a Jenkins-based CI workflow where we do all development on the dev branch, and have Jenkins automatically build and test from there before sporadically merging with master. We have written this in ./CONTRIBUTE.md, but I see that it should probably be made even clearer, and available from readme. I'll fix that. I believe some of the changes you're proposing are allready implemented on the dev-branch?   Great if you could take a look at the dev-branch version and contribute any further suggestions directly there. @AndreasAakesson please comment on this if you like.

AndreasAakesson:
I'd say the changes proposed here are already implemented in #332, but please feel free to improve the OSX build process even further :)

marssaxman:
Aha, thanks for the pointer. I am just getting started with IncludeOS and had not discovered the dev branch yet. I'll try that out and see how it goes.

PR #345, opened 2016-01-06
vbox script replaces disk so changes gets updated in VM

AndreasAakesson:
vboxrun.sh only converted the .img to .vdi if .vdi didn't exist. To update your VM with the changes from your service you had to first remove the .vdi, even if the .img (service) had changed.

The script will now remove the .vdi if it already exists, and always convert a new .vdi from .img.

### DIFF
--- a/etc/vboxrun.sh
+++ b/etc/vboxrun.sh
@@ -36,15 +36,26 @@ targetImg="$homeDir/IncludeOS/seed/My_IncludeOS_Service.vdi"
 diskname="${disk%.*}"
 targetLoc=$diskname.vdi
 
-# Remove disk if already exists
-if [[ -e $targetLoc ]]
+# Exit script if disk can not be found.
+if [[ ! -e $disk ]]
 then
-    echo -e "\n>>> Disk already exists, removing...\n"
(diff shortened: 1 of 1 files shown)


alfred-bratterud:
Hi, I'll merge this, but I'm a little concerned about removing a disk image just like that. As long as there is a .img version around there's no problem, but I guess it's possible to do `./vboxrun.sh some_completely_different_image_by_mistake.vdi` and then get that removed unintentionally. I suggest adding a test that checks whether the `$diskname.vdi` equals `$disk` (minus the path). If it does, give a prompt to have the user verify. If it's not equal we know that this is a conversion from another image, and we can overwrite. You agree?

AndreasAakesson:
It's now possible to use the script as your use case, without deleting source image.

`./vboxrun.sh my_image.vdi` will just run the disk.
`./vboxrun.sh my_image.img` will create a .vdi, and replace the current .vdi if its already there.

Also added check if input image exists, else it will exit the script.

If `$diskname.vdi` equals `$disk`, there is no need for any conversion since you already got your .vdi. Do you agree?

PR #349, opened 2016-01-18
Minor fixes to seed/Makefile and copy-scripts.sh

AndreasAakesson:
* Makefile now supports custom linker.

* Changed name from **copy-scripts.sh** to the same "name convention" as the other scripts (by using underscore). This will also fix the typo in `etc/install_from_bundle.sh`, whos currently not copying scripts.


alfred-bratterud:
Great! So now jenkins should probably work

PR #352, opened 2016-01-19
Disabled VGA output and fixed typo in vmbuilder

fwsGonzo:


fwsGonzo:
Fixes: could not initialize char* with an initializer list

PR #354, opened 2016-01-23
return proper error if fork isn't supported

raphaelsc:
fork manpage states that ENOSYS is returned if fork() isn't supported on this platform.


fwsGonzo:
Our working branch is dev, so if you create a mergable reqeuest to dev I will accept

alfred-bratterud:
btw @raphaelsc - thanks for contributing! I'd be happy to hear any thoughts you might have about IncludeOS. We're gearing up and will be increasing our activity going forward.

PR #357, opened 2016-01-29
Memdisk

fwsGonzo:
WIP filesystem implementation, with support for disks attached in service image

Note: Dont merge, its just an open pull request


fwsGonzo:
Note that you need to create a disk.sh file that starts the test_disk service
IN ADDITION you will also need a valid FAT32 or FAT16 disk image named "smalldisk" in `src` directory, for now.

PR #359, opened 2016-02-04
New install script that only downloads and installs bundle, not qemu

kybeg:
This script is needed to install the IncludeOS build environment without any runtime tools, such as qemu. This is useful for building docker images.


alfred-bratterud:
Can one of the admins verify this patch?

mnordsletten:
add to whitelist

fwsGonzo:
Looks good! Nice!

PR #366, opened 2016-02-14
inb and outb moved from kernel/os to hw/ioport

yrakcaz:
inb and outb are hardware communication things.
To be more proper, these calls are now in hw/ioport class.

### DIFF
--- a/api/hw/ioport.hpp
+++ b/api/hw/ioport.hpp
@@ -15,20 +15,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <hw/ioport.hpp>
-#include <hw/pic.hpp>
+#ifndef HW_IOPORT_HPP
+#define HW_IOPORT_HPP
 
-uint16_t PIC::irq_mask_ {0xFFFB};
+#include <common>
 
-void PIC::init() noexcept {
-  IOport::outb(master_ctrl, master_icw1);
-  IOport::outb(slave_ctrl,  slave_icw1);
-  IOport::outb(master_mask, master_icw2);
-  IOport::outb(slave_mask,  slave_icw2);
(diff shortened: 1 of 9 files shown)


alfred-bratterud:
Can one of the admins verify this patch? Comment with "ok to test" to start the build or "add to whitelist" to add a trusted developer.

alfred-bratterud:
Hi Zack, thanks for contributing! I agree with the move to hw. I'm not a 100% sure about the name `IOport` as we're trying to adhere to core guidelines which [recommends avoiding camel case](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rl-camel). But - I think this is better than having it in the OS class so I'll merge.  One suggestion: If you have time I wonder if it would actually be better to just use a namespace `hw`, and then have the functions called `hw::inb`/ `hw::outb`. @fwsGonzo, @RicoAntonioFelix @AndreasAakesson - your opinions?

yrakcaz:
Hi! I guess I can do that.
So do you think that every classes contained in the `hw/` folder should be in that new `hw` namespace?

fwsGonzo:
That sounds good. As for the inb and outb functions: They are largely inline assembly, 1-2 instructions, and don't deserve their own function calls imo. They could be inlined in your header, or as alfred wrote just (inline) functions in the hw namespace.

yrakcaz:
Great, I'll make a new pull request soon.

alfred-bratterud:
One important consideration is that we'll want support for other architectures at some point (read ARM). At that point all inline assembly has to be replaced. I guess there's no obious and pretty way to do that except by wrapping everything in macros. Any suggestions or thoughts on this, please shoot.

PR #369, opened 2016-02-15
Add a `hw` namespace

yrakcaz:
Every classes contained in the hw/ folder are now in the new `hw`
namespace, except PCI_manager (its header should probably be in
the kernel/ folder).


fwsGonzo:
Looks good!

PR #376, opened 2016-02-18
First draft of IDE device driver implementation

yrakcaz:
Hi, I just have a few questions about it :

1) For the moment, the `IDE` class only handles the `master` device. Do you
    think that it could be interesting to handle the `slave` device too?
    If so, how could we give the information to the `IDE` class?

2) I don't have implemented the `write` function yet, how should I do that?
    I saw that in the `VirtioBlk` class it's the same thing, what about it? And shouldn't
    we have an Interface `IBlkDriver` which would be implemented by `IDE`, `VirtioBlk`
    and the other block devices?

3) Could you explain me a little how should the FileSystem work in IncludeOS?
   What if I want to bind the IDE with a FS? Shouldn't we make the `hw::Disk`
   class implement the `IDiskDevice` interface?

Cheers.

### DIFF
--- a/api/hw/disk.hpp
+++ b/api/hw/disk.hpp
@@ -22,9 +22,6 @@
 
 namespace hw {
 
-/** Hopefully somebody will port a driver for this one */
-class IDE;
-
 template <typename DRIVER>
 class Disk {
 public:
--- a/api/hw/ide.hpp
+++ b/api/hw/ide.hpp
@@ -1,0 +1,73 @@
+// This file is a part of the IncludeOS unikernel - www.includeos.org
+//
+// Copyright 2015 Oslo and Akershus University College of Applied Sciences
+// and Alfred Bratterud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef HW_IDE_HPP
+#define HW_IDE_HPP
+
+#include <delegate>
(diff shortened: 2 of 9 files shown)


fwsGonzo:
Cool stuff! It's probably not interesting to handle slave too, as we are aiming at supporting things minimally but fully (if that makes sense). Most VMs will only have one harddisk, so to speak, so only supporting master makes sense here.

Btw, is your solution async or possible to make async?

fwsGonzo:
I see that its just polling. Fair enough. Want me to merge this?

yrakcaz:
Ok great!

As I wrote in my commit message, I'll probably make it async using IRQs in a next commit,
like the write function and eventually FS things.

For the merge, as you prefer, but as this first version works I think it can be merged even
if it's not totally complete.

fwsGonzo:
I'll have to make the connection between hw::Disk and the filesystem tomorrow then. I'll just merge this for now.

yrakcaz:
Ok, so nothing more needed about the IDE driver except IRQ things, and removing the `write` function?

fwsGonzo:
You wrote it, you defend it :)

Tomorrow, I'll take your IDE class and try to connect it to the filesystem. You don't have to push any changes immediately, and I won't be removing the write() proto so we don't get a conflict. We can talk more on gitter.im tomorrow if you want.

yrakcaz:
Ok for me!

alfred-bratterud:
Hi, nice work @yrakcaz, thanks :-) About `write` - good that you didn't implement ti yet!  As a general policy I we want to add all FS/Disk write-functionality as separate explicit includes. Alternatively, a minimal fs/disk that only does read, and another that does both, using the allready existing read functionality. This achieves two things: 
  * 1) we can say there is no code inside the OS that can write to the file system (we're want to make a platform suitable for [immutable infrasctructure](http://radar.oreilly.com/2015/06/an-introduction-to-immutable-infrastructure.html)) unless we explicitly asked for it
  * 2) Smaller images by default. We want small object files as a general rule, in order to get fine granularity. Say you want 10 features, and each feature has 2 unnecessary functions, you'll get 20 unnecessary functions in your image.

We can discuss the details on gitter.im tomorrow - looking forward to seeing more from you :+1:

PR #390, opened 2016-02-24
fs: FAT16 support

fwsGonzo:
Needed for smaller images


fwsGonzo:
Very troublesome to add, but we do indeed have FAT16 support now

AndreasAakesson:
:clap:

PR #403, opened 2016-03-04
test: Memdisk

fwsGonzo:


fwsGonzo:
The testdisk image looks like its just an ASCII file, but its not. Git preserves data on binary level, so its an exact copy and there also contains the quirk from the books source. At the beginning of the file there is a 0xBFBBEF byte sequence which is not text-renderable.

fwsGonzo:
For the disk out of bounds test I had to replace the asserts that verified read position in memdisk with returning a 0x0 shared pointer, otherwise its impossible to successfully continue the test.

Issue #221, opened 2015-11-04
TCP::Socket implement connect


alfred-bratterud:
We've only got passive connect (TCP::Socket::onConnect) as of v0.6.3

AndreasAakesson:
#370 needs testing, been unable to do so in my current environment. Will update when tested.

alfred-bratterud:
Fixed by @AndreasAakesson somewhere before #407

Issue #220, opened 2015-11-04
Implement TCP active close 


alfred-bratterud:
Currently we've only got passive close

AndreasAakesson:
#370 can active close a connection. Not yet fully tested, but implementation should support it.

alfred-bratterud:
Fixed by @AndreasAakesson somewhere before #407

PR #407, opened 2016-03-07
test: TCP more work

AndreasAakesson:


AndreasAakesson:
Time to bury #220 #221 :skull:

PR #426, opened 2016-03-12
Serial port with on_data / readline events

alfred-bratterud:


alfred-bratterud:
Tested on hardware (trident1) and in virtualbox. @fwsGonzo @AndreasAakesson Let me know if this causes any trouble for you, otherwise I'll merge.

Issue #417, opened 2016-03-10
Implement TCP Options [RFC 793]


AndreasAakesson:
```
      Kind     Length    Meaning
      ----     ------    -------
       0         -       End of option list.
       1         -       No-Operation.
       2         4       Maximum Segment Size.
```

This includes actually having support to parse and set option, as the current solution don't.

Broken out from #408.

fwsGonzo:
Good stuff :+1:

alfred-bratterud:
Nice!

PR #428, opened 2016-03-13
tcp: Added header options support

AndreasAakesson:
TCP can now parse and add options.

`test/tcp` passed locally.

Solves #417.


AndreasAakesson:
@fwsGonzo fixed! thanks for the help :camel:

PR #434, opened 2016-03-14
hw/ide: asynchronous reading now works properly

yrakcaz:
* The first commit adds these changes
* The second one updates the tests

### DIFF
--- a/api/hw/ide.hpp
+++ b/api/hw/ide.hpp
@@ -19,6 +19,7 @@
 #define HW_IDE_HPP
 
 #include <delegate>
+#include <list>
 
 #include "disk.hpp"
 #include "pci_device.hpp"
@@ -30,8 +31,8 @@ class IDE : public IDiskDevice {
 public: 
   enum selector_t
   {
-    MASTER,
-    SLAVE
+    MASTER = 0x00,
+    SLAVE = 0x10
   };
   
   /**
@@ -58,17 +59,18 @@ public:
   virtual block_t size() const noexcept override
   { return _nb_blk; }
 
+  static void set_irq_mode(const bool on) noexcept;
+
+  static void wait_status_busy() noexcept;
+  static void wait_status_flags(const int flags, const bool set) noexcept;
+
 private:
   void set_drive(const uint8_t drive) const noexcept;
   void set_nbsectors(const uint8_t cnt) const noexcept;
   void set_blocknum(block_t blk) const noexcept;
   void set_command(const uint16_t command) const noexcept;
-  void set_irq_mode(const bool on) const noexcept;
 
-  void wait_status_busy() const noexcept;
-  void wait_status_flags(const int flags, const bool set) const noexcept;
-
-  void irq_handler();
+  void callback_wrapper();
   void enable_irq_handler();
 
 private:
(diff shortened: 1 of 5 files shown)


alfred-bratterud:
Hi @yrakcaz thanks for another contribution! 
There are some issues with your solution:  When you [call the subscription callback](https://github.com/yrakcaz/IncludeOS/blob/dev-branch/src/hw/ide.cpp#L243) you're in a preempted context that hasn't been properly saved. The default IRQ-entry does save general purpose registers, but that's it, and it's not enough. What if the service was in the middle of a series of floating point operations? The point is that you can't know what a user callback will do, so you can't take any chances. I suggest a small rewrite:
  1. Go back to using the IRQ_manager's event system to assign a subscription, but only to handle the callback
  2. Keep your IDE irq-handler, but make it *not call the callback*, but simply increment the IRQ, just like the normal irq-handlers. That way the event handler can call the callback when it has time - in a non-preempted context. I've now made it easier for you to "do the right thing", by wrapping the default irq-handling into a function, [`IRQ_manager::register_interrupt(int irq)`](https://github.com/hioa-cs/IncludeOS/blob/78d1fe336f801cdcdc9b15b1a6e337f5bd594c9b/api/kernel/irq_manager.hpp#L153) (78d1fe336f801cdcdc9b15b1a6e337f5bd594c9b). You can simply call that function to have the event system know about the event, and trigger the callback in an orderly fashion later.

You can still do the buffering inside the IRQ-handler, if it's necessary, i.e. that the device has to be serviced directly after the interrupt. Is this the case?  The point is that you now have a way to do the time-sensitive part in the IRQ-handler directly, but you should still defer as much work as possible to the event loop, since that's what will be running in a non-preempted context.

yrakcaz:
Hi,

Thanks a lot for your quick response. 

Effectively I didn't thought about that, I'll try to look at this problem and make à new PR as soon as possible.

Cheers.

yrakcaz:
Hi,

I just wrote a new version : think this one is better.. Does it seems consistent to you?

PS : Sorry for the wait, I wasn't available for a few weeks.

Cheers.

alfred-bratterud:
Hi @yrakcaz I've looked a bit more at your code and here are a few comments:
1. Your test needs to use `CHECK`/`CHECKSERT` so that it's possible to determine if it passes or not. See e.g. the `bufstore`-test
2. How do you interface with a secondary hard drive? I'd like to see that in the test as well.
3. I didn't really understand why you needed to write your own interrupt handler (I just trusted that you do), but now that you have I'd like to be absolutely sure that you don't mess up the preempted context. For example if you started doing a long series of floating point operations immediately after calling async read, you should be able to verify that your interrupt handler (or any functions it calls) doesn't clobber floating point registers. 

You don't have to have all of this in order for me to merge, as the dev branch is supposed to represent WIP. But 1 is a requirement otherwise we can't automate testing.

alfred-bratterud:
(Of course your test could also use the `lest` unit-testing framework like in the `./test/STL` if you prefer )

alfred-bratterud:
@yrakcaz, I'll merge now and create an issue instead, assigning it to you. This way we'll save you the trouble of dealing with merge conflicts due to the impending Nuke ( #462 )

fwsGonzo:
The CPU registers all need to be saved in order to execute arbitrary in that IRQ handler, so right now all the test shows is that the IRQ runs. But it won't show that the OS is not sane after running the handler.

PR #437, opened 2016-03-15
Minor changes to TCP module

RicoAntonioFelix:
## Changes
* Refactored the `TCP::Socket` class

NOTE: Whitespace changes due to editor settings...

### DIFF
--- a/api/net/tcp.hpp
+++ b/api/net/tcp.hpp
@@ -56,63 +56,90 @@ public:
 
 public:
 	/*
-		An IP address and a Port.
-	*/
-	class Socket {
-	public:
-		/*
-			Intialize an empty socket.
-		*/
-		inline Socket() : address_(), port_(0) { address_.whole = 0; };
+	 * @type TCP::Socket
+   	 * @description An IP address and port pair <ipaddress:port>
+   	 */
+  	class Socket {
+  	public:
+    /*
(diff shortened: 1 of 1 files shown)


alfred-bratterud:
@RicoAntonioFelix are all the whitespace changes caused by editorconfig? We should try to do one separate commit that applies the editorconfig to all files in one go.

RicoAntonioFelix:
@alfred-bratterud Yes I noticed formatting changes in my editor due to the .editorconfig file...

alfred-bratterud:
Hi @RicoAntonioFelix, sorry about the delay. I'd like to merge - but here's a couple of formatting details that really aren't important, but since this PR is about formatting anyway well: 
  1. we prefer to not use the multiline-comment style with a `*` on each line. It just makes the comment much harder to edit. 
  2. What is the significance of the `@type` and `@description` decorations? Are they for doxygen? They look like it, so if they are that's fine, if not we should make sure there's no conflict.

alfred-bratterud:
@RicoAntonioFelix I'll close this one off, since it mostly does whitespace changes, which I'm now doing in batch-mode in #462. Please feel free to reopen a PR after 15:00 CEST today, when we have applied all the whitespace changes, just keep in mind my previous comments.

PR #446, opened 2016-03-17
Pic revamp

alfred-bratterud:
The PIC-class should now better reflect the Intel 8259A / 8259A-2 1988 whitepaper


alfred-bratterud:
@fwsGonzo and @AndreasAakesson please comment - in particular you'll notice 145e44d

alfred-bratterud:
Ok, so the RFC timeout expired on this one. Merging.

PR #493, opened 2016-04-16
Update api/common

RicoAntonioFelix:
## Changes
* Redefined termination policy for Expects and Ensures...

### DIFF
--- a/api/common
+++ b/api/common
@@ -45,7 +45,6 @@
 #  define UNUSED(x) UNUSED_ ## x
 #endif
 
-
 /* BOCHS Break point */
 #define BREAK __asm__ volatile("xchg %bx,%bx");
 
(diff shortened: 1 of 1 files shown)


fwsGonzo:
Cool :+1:

alfred-bratterud:
Looks good @RicoAntonioFelix . Question: why the while-loops?

fwsGonzo:
the while loops act as scope-creation helper, just like FreeBasics `scope` keyword:

http://www.freebasic.net/wiki/wikka.php?wakka=KeyPgScope

alfred-bratterud:
And another thing: @RicoAntonioFelix - I was thinking we should probably use our own macro for enabling / disabling this mode, something like : `#ifdef ASSERT_ON_CONTRACT_VIOLATION`. This way we can keep the GSL versions completely unchanged. Thoughts?

RicoAntonioFelix:
@alfred-bratterud Check out the updates from your suggestions...

PR #494, opened 2016-04-18
install: Add git submodule update

fwsGonzo:


alfred-bratterud:
Great - this is needed in `install.sh` as well

PR #496, opened 2016-04-19
[Ethernet] Replaced all assert statements for Expects...

RicoAntonioFelix:


alfred-bratterud:
Hi, this is nice, just remember that `./api` is allready on the include path, so the correct include statement for `common` is simply `#include <common>`. Also the reason why we take a while to respond to your PR's is simply a lack of automated testing. That's likely to be right around the corner though- @mnordsletten set up a brand new jenkins yesterday.

PR #499, opened 2016-04-20
virtioblk: Queue for requests

fwsGonzo:


fwsGonzo:
virtioblk should be more or less done now
in the future we may need SCSI to really bump up the speed

PR #502, opened 2016-04-22
Updated an environment variable in etc/copy_scripts.sh

mnordsletten:
Will now properly support installation in a different directory than $HOME.


alfred-bratterud:
Nice. That substitution syntax was new to me - we can use this to simplify the install scripts a lot!

fwsGonzo:
Hopefully the same method works on macs too

mnordsletten:
The same syntax has been present in this file for a long time, so this has been in use since January I think. Don't know when the last time someone tested on a mac though.

PR #512, opened 2016-05-01
Augmented filesystem components...

RicoAntonioFelix:
## Changes
* Added some API documentation to the various common components of the filesystem
* Removed `buffer_t` declaration from `filesystem.hpp` as its already defined in the included `common.hpp` file
* Augmented method `Buffer::is_valid()` to incorporate a check for the state of the `error_t` object
* Added license block to `filesystem.cpp` file
* Made `no_error` const as its value is fixed

### DIFF
--- a/api/fs/common.hpp
+++ b/api/fs/common.hpp
@@ -6,9 +6,9 @@
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


fwsGonzo:
Looks good, although quite a bit of formatting changes

alfred-bratterud:
Hi @RicoAntonioFelix thanks for the documentation! I also noticed that you comments are now doxygen-friendly which is great. However, we'd like to use a consistent commenting style throughout, and we just had a little discussion, and did a few tests with doxygen. Our suggestion is the following policy:

- Multiline comments, to document classes / functions, *in header files* should use this style:
```
   /**                                                                                                                           
     * This type of  mutiline comments.                                                                                           
     *                                                                                                                            
     * @param param1 : This shows up nicely in doxygen without too                                                                
     * much vertical space (scrolling)                                                                                            
     * @param param2 : We don't need newlines here. Colons work.                                                                  
     *                                                                                                                            
     * @warning This is a warning, that doxygen nicely marks                                                                      
     * red. As long as this section is separated by newlines, the                                                                 
     * whole paragraph comes up as a warning.                                                                                     
     *                                                                                                                            
     * @todo Doxygen can auto-generate a TODO-list based on these                                                                 
     * entries. Again separating the section with newlines marks it as                                                            
     * a TODO.                                                                                                                    
     *                                                                                                                            
     * A more detailed descripiton can follow. Doxygen will                                                                       
     * automatically make the first sentence into a "brief"                                                                       
     * description. Gunzo likes the double stars at the start and the                                                             
     * end, so we'll go with that.                                                                                                
     **/
void my_excellently_documented_function(T1 param1, T2 param2);
```

In Doxygen this comes up like so:
image


- One-liner comments: According to the core gudelines, comments should express intent. There's no need to comment things that should be obvious to an average programmer, but complex stuff can be useful to comment briefly like:
```
// Compute the square root
x *= x;
```
...Now the comment helps the reader understand what I intended (which obviously isn't what the code is actually doing)

Apart from this I don't think there's any need for extraneous comments. @fwsGonzo @AndreasAakesson @RicoAntonioFelix can we agree on this or do you have comments?

fwsGonzo:
I made some necessary changes to fs, unfortunately, wish I had merged your changes earlier
I could integrate them myself, but then it won't show you as having made the changes

RicoAntonioFelix:
The generated documentation looks good, so I have no problem with the commenting style...

@fwsGonzo No problem... I'll close this one and start over...

Since I can't do any testing of the system on my current machine which implies I can't do any major changes or additions, I decided to do documentation for the while to be at least productive... (Still researching `MongoDB`)

PR #521, opened 2016-05-10
tcp: now renewing instead of reseting read buffer

AndreasAakesson:
Changed to rather "renew" (allocating a new) buffer instead of reseting the current one. Cost between `clear` (memset) and `renew` (make_shared) maybe is the same? It also makes Connection release the ownership of the buffer when giving it to the user via user callback.


AndreasAakesson:
Added a hotfix.

PR #535, opened 2016-05-22
Newlib / binutils update

alfred-bratterud:
Updated newlib to latest version

### DIFF
--- a/api/sys/features.h
+++ b/api/sys/features.h
@@ -6,9 +6,9 @@
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


alfred-bratterud:
@mnordsletten these tests fail because I've updated IncludeOS to depend on a newer version of newlib. This is a catch 22 for us; we can't update the binary bundle Jenkins uses unless we do a release, and we don't want to do a release before Jenkins can verify that everything still works.

alfred-bratterud:
I'll do a release candidate with a new bundle and work from there.

mnordsletten:
@alfred-bratterud Allright. I was wondering what was up. I'll check back later/tomorrow to see that everything is working.

Issue #545, opened 2016-05-31
Timer test fails


alfred-bratterud:
After the new APIC / MSI-x pr #543  was merged `./test/timers` fails just when the last timer is supposed to trigger, due to a CPU exception:

```
1s. PULSE #25 
                [x] After 25 sec, 1s timer x 30 == 25 times, 2s timer x 15 == 10 times
                [x] There are now 2 timers left

ISR: 0x0  IRR: 0x0

[0] Return 0x2091ff
[1] Return 0x26441e
[2] Return 0x2035bd
[3] Return 0x2021ec
[4] Return 0x202045
[5] Return 0x201ce1
[6] Return 0x464c457f
>>>> !!! CPU EXCEPTION !!! <<<<
Heap end: 0x2d3000 

```

alfred-bratterud:
I'm getting the same result consistently on Virtualbox and bare metal KVM

alfred-bratterud:
Setting the Qemu `-smp` parameter to 1 removes the problem as well.

alfred-bratterud:
Verified on both bare metal KVM, VirtualBox and nested in openStack (Jenkins).

alfred-bratterud:
Fixed in dev after #578. I'll consider hotfixing master with this. As this only applies to stopping repeated timers, which isn't used anywhere in the core system, it shouldn't cause instability unless your service stops a repeated timer.

PR #546, opened 2016-05-31
Wrapped the IRQ_handler instance (bsp_idt) in static singleton. 

alfred-bratterud:
Might solve #545. @fwsGonzo you'll probably have opinions on how I name this one. As you can see I've added a numeric argument in case we'd want to use the getter for several instances. I'm not sure we want to expose a getter for each CPU though - I think it would probably be better to have a static function, or some other global IRQ supervisor that could assign your IRQ to whichever CPU was in charge of that particular line (I know they use IRQ balancing in Linux by default) 

However, we don't need to pollute the global namespace, plus avoiding using global constructor for the `bsp_idt`-instance seems to solve #545 for me on trident3.

### DIFF
--- a/api/kernel/irq_manager.hpp
+++ b/api/kernel/irq_manager.hpp
@@ -95,10 +95,9 @@ public:
 
   /**
    *  Subscribe to an IRQ
-
+   *
    *  @param irq: The IRQ to subscribe to
    *  @param del: A delegate to attach to the IRQ DPC-system
-
    *  The delegate will be called a.s.a.p. after @param irq gets triggered
    *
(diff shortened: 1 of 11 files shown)


alfred-bratterud:
Oh, and white-space alert. Looks like you're not using editorconfig anymore?

alfred-bratterud:
All checks pass so Jenkins agree. I'll merge in order to get the red build-badge back to green, and we can discuss the details / further changes tomorrow.

fwsGonzo:
Looks good to me, just wondering about the IDs and how they relate to numactl stuff. Ideally cpu(0) is always BSP but I don't know

Issue #547, opened 2016-06-02
FAT::readFile causing CPU exception/reading outside RAM


AndreasAakesson:
Either qemu crashes, or CPU exception occurs on `FAT::readFile(const string&, on_read_func)`, when using **VirtioBlk** FAT16 disk.

fwsGonzo:
Recommend making readFile inline using stat() and read() instead of it being virtual interface function.

AndreasAakesson:
:+1:

PR #550, opened 2016-06-03
Refactored Filesystem::readFile

AndreasAakesson:
Now uses inline `stat() => read()`. Also renamed to `read_file`.


AndreasAakesson:
Should solve #547

PR #562, opened 2016-06-07
Fixes to ssp, vbox and show addr for missing symbols on backtrace

fwsGonzo:


alfred-bratterud:
Cool - would actually be nice to have the address printed for all the stack frames, as well as the name :-) I'll merge once I'm done with the cherry-picking stuff

PR #564, opened 2016-06-07
Release 0.8.0

alfred-bratterud:
There are still issues with building the latest llvm from source, but install from bundle (default) seems to work on all supported platforms

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,22 @@
 ![IncludeOS](./doc/IncludeOS_logo.png)
 ================================================
-IncludeOS is a [Unikernel](https://en.wikipedia.org/wiki/Unikernel) written from scratch in C++, designed for x86 hardware virtualization, with no dependencies except for the virtual hardware. [Read more on the wiki](https://github.com/hioa-cs/IncludeOS/wiki). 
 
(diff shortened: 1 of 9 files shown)


alfred-bratterud:
Some issues to work out:
* Looks like jenkins expects all tests in master to be just like in dev. So either we update all the tests in master or we change the master build-jobs in jenkins. @mnordsletten - is this only for the PR-tests or will this make the master badge go red? 
* TCP test times out. Was this @AndreasAakesson was this fixed in dev?

mnordsletten:
Yeah at the moment the PR tests that you can see were performed here expect that the tests are just like in the dev branch. This is something I will have to work out. 

When you eventually merge dev into master the exact same tests should be completed in order to update the badge. This uses the ./test/validate_all.sh script to check which of the tests are ready to be tested. So if the dev badge is OK before merge the master badge should be OK after the merge. For both of these badge-tests there is an option in jenkins to disable certain tests. So at the moment the transmit test has been disabled for both master and dev.

As for the tcp test, this was increased in the dev branch. We did still see if time out a couple times after this, but checking the badge_test history it's been a long time since this was an issue now.

PR #566, opened 2016-06-08
Release 0.8.0 (take two)

alfred-bratterud:
LLVM issues should now be resolved. The tests that Jenkins uses for the master pipeline also passes on my clean 14.04 test machine, so in theory we should be all set (but there are ghosts in the machine)


alfred-bratterud:
Catching! All tests pass.

Issue #565, opened 2016-06-08
TCP interprets ethernet padding as data


alfred-bratterud:
TCP calculates the size of the data in a packet as `net::Packet::size()` minus the size of all headers. This means that ethernet padding will be interpreted as data. The correct behavior is to ask the IP header for IP data size, i.e. IP segment length minus size of IP header, and then subtract the size of the TCP header.  (Thanks to @sfrode at Basefarm for the packet trace that pointed us in this direction)

AndreasAakesson:
I'm on it!

AndreasAakesson:
This is probably solved

PR #569, opened 2016-06-09
Refactored PacketIP4 and TCP::Packet

AndreasAakesson:
* Made size/length calculation more dynamic
 * TCP Packet is now using functions inside PacketIP4 to determine offsets.
* Added prefix to a lot of functions.

Hopefully solves #565 (earlier failed scenario is now working).

Also made demo only reply with HTML on `GET / ` requests.


AndreasAakesson:
test

mnordsletten:
Ok to test

AndreasAakesson:
Retest this please

PR #570, opened 2016-06-10
TCP hotfix

AndreasAakesson:
TCP Packet now correctly calculates it length with help from IP Packet.

Fixes bug in #565


AndreasAakesson:
Retest this please

PR #571, opened 2016-06-10
Updates to install.sh and install_from_bundle.sh

mnordsletten:
Added shebang to install.sh, pull requests should work with jenkins now.
Also removed the oauth token from install_from_bundle.sh and added some extra output.


mnordsletten:
Retest this please

mnordsletten:
It now pulls the correct bundle.

PR #572, opened 2016-06-10
New install scripts

mnordsletten:
Moved all bundle download code and IncludeOS install commands to the now common script etc/install_from_bundle.sh. Removed all ubuntu specific code from this script and placed this in a seperate script called etc/install_ubuntu.sh. 

Also updated some environment variables and the way they are declared.


mnordsletten:
Retest

mnordsletten:
Retest this please

PR #585, opened 2016-06-17
tcp: Clean up (possible) copies of shared_ptr in callbacks

AndreasAakesson:
Last step is `TCP` cleaning up a Connection (closing). It removes the `Connection_ptr` from it's map, so the last pointer goes out of scope, destroying the Connection and releasing all it's resources.
Due to Connection_ptr being copied into lambdas, the Connection always had one (or more, depending on how many callbacks copying ptrs) instance of itself, making it invincible (the bad kind).

```
conn->read(1024, [conn] ... // copies here
```
Connection now calls `reset()` on all `delegate` (callbacks) right before closing.

### DIFF
--- a/api/net/tcp.hpp
+++ b/api/net/tcp.hpp
@@ -610,6 +610,10 @@ namespace net {
         ReadRequest(size_t n = 0) :
           buffer(buffer_t(new uint8_t[n], std::default_delete<uint8_t[]>()), n),
           callback([](auto, auto){}) {}
+
+        void clean_up() {
+          callback.reset();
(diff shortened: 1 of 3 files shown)


alfred-bratterud:
Nice:-)

AndreasAakesson:
ok to test

alfred-bratterud:
Looks like this did the trick :-) Master still has some stability issues under heavy bombardment but it's not a memory leak anymore. Both the bug and the solution was interesting, so that's officially a 🎓 :-)

PR #591, opened 2016-06-21
Possible fix to stress + tcp test failing

mnordsletten:
Added a simple wait function to the stress/test.py which makes sure it does not continue until the TIME_WAIT connections are below a set limit. Not sure if this is a solution you want, but it makes sure this test does not interfere with other tests.

Last commit fixes some irregular function calls.


mnordsletten:
Now passes all tests in the pipeline on my personal fork: https://jenkins.includeos.org/job/personal_pipeline_mnordsletten_dev/

PR #601, opened 2016-06-23
fs: fixed circular dependency in async next-loop (master hotfix)

### LINKED ISSUE #602: fs: async next-loop causing circular dependency
[`shared_ptr<function<void()>>` consists of lambda with a capture of itself](https://github.com/hioa-cs/IncludeOS/blob/master/src/fs/fat_async.cpp#L30) (cyclic reference)
```
auto next = std::make_shared<next_func_t> ();
*next = 
[this, sector, callback, dirents, next] (uint32_t sector)
...
```
This is causing the function to never get released. The function also captures `callback`, which may contain other larger objects, resulting in keeping them from deallocating as well.

This is similar to what caused #583.

AndreasAakesson:
Fixes #602

### DIFF
--- a/src/fs/fat_async.cpp
+++ b/src/fs/fat_async.cpp
@@ -11,7 +11,7 @@
 #define likely(x)       __builtin_expect(!!(x), 1)
 #define unlikely(x)     __builtin_expect(!!(x), 0)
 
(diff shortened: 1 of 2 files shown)


AndreasAakesson:
Not sure why this commit includes a lot of whitespace changes, but #600 has none..

Issue #473, opened 2016-04-08
Packet and Buffer: Merged


fwsGonzo:
By preallocating a memory pool of buffers with size `sizeof(Packet) + buffer_size`, we can merge packets and buffers together for increased memory locality. We can also access the buffer of a Packet without a memory lookup: `(char*) this + sizeof(Packet)`. This is likely to be very efficient compared to the current solution. Since Packets are just metadata headers for buffers anyway, we can also use VLA to solve this without having to do the calculation ourselves:
```
struct Packet
{
    size_t len;
    IP4::addr next_hop;
    // VLA:
    char buffer[0];
}
```

We no longer have to allocate Packets when we use `createPacket` from inet. Instead we will just be using shared_ptr with a custom deleter for Packet. shared_ptrs are stack-stored.

AndreasAakesson:
:clap:

fwsGonzo:
while working on this, we can fix https://github.com/fwsGonzo/IncludeOS/commit/ef3cb11bebecc52dccec2478597becee16d56661

fwsGonzo:
Work started with https://github.com/hioa-cs/IncludeOS/commit/04830a756e39d9c270ae3f40c769e7fe6383a336

fwsGonzo:
This has been implemented, and is working nicely

PR #609, opened 2016-06-24
net: Allocate packet and data in same buffer

fwsGonzo:
Increased memory locality etc.


fwsGonzo:
This is the implementation for #473

fwsGonzo:
Need to update bufferstore first, then update the bufferstore test for the bufstore test to pass

PR #612, opened 2016-06-26
Added functionality to easily bring up an IPv4 networking stack...

RicoAntonioFelix:


fwsGonzo:
I dont think it makes sense to have defaults for the network config - when it's different for every place where the OS will be used. Those values are only valid when running our very specific testing setup with the tap-device. That said, most places will have dhcp running. I say let's try without defaulting IP, netmask and gateway. Otherwise looks good.

RicoAntonioFelix:
Guess this should be okay now based on your feedback...

fwsGonzo:
ok to test

Issue #615, opened 2016-06-27
v0.8.1 tag now points to merge commit, only existing on master branch


mnordsletten:
With the hotfixes introduced to move us up to v0.8.1 the latest tag was linked to the commit of the merge on the master branch.

The tag now points to: https://github.com/hioa-cs/IncludeOS/commit/5bab8cec6a7841e021bb30eb3c037dbadc731b66 which only exists on the master branch. This then makes the `git describe --abbrev=0` command from `etc/install_from_bundle.sh` fail as the commit for v0.8.1 does not exist.
This issue can be easily overcome with `git describe master --abbrev=0`, but a more robust solution that does not rely on the tags of your local repos is wanted. We had lots of issues with outdated tags on personal forks, so a solution that utilizes `git ls-remote` is preferred. We must try to avoid a _hack_ solution though. 

I will try to find a good universal solution to this issue.

PR #618, opened 2016-06-27
A fix for pulling down tags

mnordsletten:
Does not rely on tags at all any more. Will now rather find the latest release from the json output from [releases](https://api.github.com/repos/hioa-cs/IncludeOS/releases). Small change to the python script which now let's you extract multiple fields from the json output. Fix for #615

### DIFF
--- a/etc/get_latest_binary_bundle_asset.py
+++ b/etc/get_latest_binary_bundle_asset.py
@@ -1,12 +1,21 @@
 #!/usr/bin/python
+""" Will return one of fields of the JSON output supplied. Expects to get two
+    command line inputs:
+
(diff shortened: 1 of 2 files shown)


AndreasAakesson:
This looks like a good general solution imo.

mnordsletten:
The only setback to this is if the json output for some reason one day is not in the order with the latest tag first. 

When we make a proper release (and not pre-releases) we can use the shortcut https://github.com/hioa-cs/IncludeOS/releases/latest and always get the latest tag. This does not work with prerelases though. This might need some modification at one point when we eventually make a full release, but this feels more robust right now.

PR #625, opened 2016-06-29
Removed sudo calls from vmrunner script

mnordsletten:
Changes to the sudo call in ubuntu 16.04 made the old way of calling the qemu process with sudo not work.
Now displays error if vmrunner.py is not called with sudo.

### DIFF
--- a/test/vmrunner.py
+++ b/test/vmrunner.py
@@ -184,7 +184,7 @@ class qemu(hypervisor):
     if self._config.has_key("mem"):
       mem_arg = ["-m",str(self._config["mem"])]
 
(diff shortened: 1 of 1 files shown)


mnordsletten:
ok to test

mnordsletten:
Ok to test

mnordsletten:
Ok to test

mnordsletten:
A full explanation of the issue we had here can be found here: http://stackoverflow.com/questions/34337840/cant-terminate-a-sudo-process-created-with-python-in-ubuntu-15-10

PR #634, opened 2016-06-30
Improved gettimeofday

ingve:
Not a final solution, but better than being back in 1970 all the time.

The snippet 

std::string make_daytime_string() {
    time_t now = time(0);
    return ctime(&now);
}

now returns "Thu Jun 30 14:37:52 2016" when cmos returns "2016-06-30T14:37:52Z". 

Previously ctime returned a time and date on Jan 1st 1970.

### DIFF
--- a/src/kernel/syscalls.cpp
+++ b/src/kernel/syscalls.cpp
@@ -24,6 +24,7 @@
 
 #include <os>
 #include <kernel/syscalls.hpp>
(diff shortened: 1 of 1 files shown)


fwsGonzo:
add to whitelist

mnordsletten:
Looks like Jenkins is commenting once per pull request job. I'll see if this is something that can be fixed.

PR #637, opened 2016-07-01
kernel: Work on heap placement, bufferstore, elf parser, profiler output

fwsGonzo:


fwsGonzo:
Turns out the separated MSI-X handlers caused some kind of deadlock in the stress test. I reverted it.

PR #645, opened 2016-07-03
elf: Heap-safe backtrace can survive OOM

fwsGonzo:


fwsGonzo:
It's also nice to know that backtrace kinda works with stripped, just no function names. Exceptions work normally, and all other strings are present.

PR #670, opened 2016-07-05
Fix to tests

AndreasAakesson:
Don't merge yet.


AndreasAakesson:
@mnordsletten Removed quick_exit from Util tests

Issue #667, opened 2016-07-05
Purge all camelCases


AndreasAakesson:
🐫 👎 💀 

Replace API-calls using camelCase with snake_case

🐍 👍 👌

AndreasAakesson:
Let's say this is done. Could probably be some more places but hard to track down.

Issue #659, opened 2016-07-05
test: serial incomplete


AndreasAakesson:
- Missing new test system (test.py)
- Missing README

This test does nothing. Any ideas how to make it do more than nothing?

fwsGonzo:
Needs to be connected to a pseudo terminal so we can read and write to the running service

alfred-bratterud:
Vmrunner has control over both stdin and stdut of the process so we could implement something like vm.writeline(str) in vmrunner.py and use that.

-------- Original Message --------
Subject: Re: [hioa-cs/IncludeOS] test: serial incomplete (#659)
From: Alf-André Walla
To: hioa-cs/IncludeOS
CC:


Needs to be connected to a pseudo terminal so we can read and write to the running service

-

AndreasAakesson:
I'm currently trying to write to qemu from python (using `_proc.stdin.write()` on the qemu process), but it doesn't look like the service (IncludeOS) receives any interrupts. Can someone confirm that it should receive interrupts? From what I can understand, `-nographic` makes serial port default to `stdio`.
Is it something I'm missing?

AndreasAakesson:
After a lot of digging I got it working.

Added `IRQ_manager::cpu(0).enable_irq(4);` in the service and now it's working. This should be done inside `Serial` when subscribing?

AndreasAakesson:
Nvm, I'm wrong. It looks like writing to subprocess (qemu) stdin just echos the text back to stdout.. So I still need to find out how to write to serial port from outside qemu.

fwsGonzo:
if serial uses legacy IRQs then yes, add enable to serial constructor

PR #683, opened 2016-07-13
Major TCP refactoring + changes to test

AndreasAakesson:
# TCP
A lot of refactoring, changes to the API and some improvements to withstand SYN flood.

## Refactoring
* No more nested classes inside `TCP` - this is now put inside the namespace `net::tcp`.
* Everything tcp related is now moved one level deeper into `/api/net/tcp/`.
* Gave (almost) every class/struct their own header.
* Reduced the amount of dependencies - especially from `class TCP`.
 * Added a very useful `common.hpp` - protocol related stuff is found here instead of `TCP`

## API changes
* New class `tcp::Listener` - this is now what you get when calling `TCP::bind`.
 * Handles every incoming connection attempt - keep tracks of all it's [half-open connections](https://en.wikipedia.org/wiki/TCP_half-open).
 * Releases the `Connection_ptr` to `TCP` when the connection has been established.
 * Withstands SYN flooding - current limit is set to `64` half-open connections. When queue is full and a new attempt is received, oldest half-open connection gets dropped.
 * Hopefully makes the API clearer; It's no longer possible to read, write, or add any other events to a listener, other than `on_accept(tcp::Socket)` & `on_connect(tcp::Connection_ptr)`.
* Replaced all(?) camelCased API calls with snake_case; `onDisconnect` => `on_disconnect`. Work towards #667.
* `Connection` now has a private callback `_on_cleanup`, used by `TCP`/`Listener` to remove their copy of the shared_ptr when the connection is shutting down. No more explicit call to `TCP::close_connection`

Every callback assignment to a connection should now be set inside `Listener::on_connect`. To see the difference, checkout `/examples/demo_service` and `/test/tcp`

# Test
Does not really solve any issues. Tried to make vmrunner write to qemu with `stdin`, but without success (#659). Also tweaked the FAT tests.

# Other
Enabled IRQ in `hw::Serial` and doubled amount of buffers in `VirtioNet`'s bufferstore.
Added `Inet4::negotiate_dhcp` (#684)

Most of the tests and examples will break with the new changes to the API. Making a pull request to get an overview of which tests/examples to update.


AndreasAakesson:
I plan on changing the signature for some of the callbacks (removing `Connection_ptr` as argument), and also rework `Connection::read` to `Connection::on_read` (or equal), since this one is a bit confusing.

AndreasAakesson:
Up to this point seems stable. Been running the TCP test several times on the same instance, and also tried SYN flooding.
Been working some more on changes to the API, and a lot of reordering/restructuring to make the headers easier to read. After this, the above case is not true, since memory gets corrupted.
Progress can be seen here: https://github.com/AndreasAakesson/IncludeOS/tree/tcp-refactor [851a566f209a3d708ac7285a4f08a8598cd2b0ea](https://github.com/AndreasAakesson/IncludeOS/commit/851a566f209a3d708ac7285a4f08a8598cd2b0ea)

alfred-bratterud:
Looks good 👍

Issue #687, opened 2016-07-27
Constructing IP-address from string


alfred-bratterud:
There should be a constructor to create an IP4-address from a string.  Very easy fix, just needs doing.

fwsGonzo:
Fixed with https://github.com/hioa-cs/IncludeOS/pull/692

PR #690, opened 2016-07-29
ipv4.hpp: Added constructor to create an IPv4 address from a std::string

RicoAntonioFelix:


alfred-bratterud:
Hi @RicoAntonioFelix great - looks like this will take care of #687. We really need a test though - would you mind cooking up a small LEST unit test for the IP class? It could be put under ./test/ip.

RicoAntonioFelix:
Hey @alfred-bratterud Sure no problem :+1:

fwsGonzo:
:+1:

PR #692, opened 2016-07-31
ethernet.hpp: Fix Ethernet MAC address string representation to print leading zeros 

RicoAntonioFelix:
# Change
* Fix Ethernet MAC address string representation to print leading zeros
* Added test folder for net::Ethernet module

### DIFF
--- a/api/net/ethernet.hpp
+++ b/api/net/ethernet.hpp
@@ -74,7 +74,7 @@ namespace net {
       // hex string representation
       std::string str() const {
         char eth_addr[18];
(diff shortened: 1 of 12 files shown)


fwsGonzo:
ok to test

fwsGonzo:
Looks like a minor change to add constructors to Ethernet addr is in order, to avoid double braces

RicoAntonioFelix:
Should I also fix the IP4 errors from here as it runs with the test?

fwsGonzo:
All tests have to pass

PR #695, opened 2016-08-05
Fixes to timers, add timer system test

fwsGonzo:


fwsGonzo:
Apparently Jenkins runs on ancient ubuntu, which has clang3.6 which has the `auto`-lambda problems.

PR #698, opened 2016-08-09
Multiboot and memory map

alfred-bratterud:
The implementation is rough around the edges and touches lots of low-level stuff so expect trouble.


alfred-bratterud:
Oups, should go to dev

PR #699, opened 2016-08-09
Multiboot and memory map

alfred-bratterud:
The implementation is rough around the edges and touches lots of low-level stuff so expect trouble.


alfred-bratterud:
I told you to expect trouble :-D Looks like the vmbuilder doesn't get built due to GSL dependencies. It's trying to use the GNU standard C++ library (libstdc++) which seems to be too old (or something). I can try to change it to use LLVM's libc++ and hope that it's up-to-date, but sooner or later we'll need to deprecate clang-3.6 and consequently ubuntu 14.04.

alfred-bratterud:
Update: vmbuilder builds after installing gcc-5. To do this we need:
```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt-get update
sudo apt-get install gcc-5 g++-5
````
The simplest solution is probably to add these lines to the install script in case of ubuntu < 16.04. However, it's pretty strange to install two compilers.  @mnordsletten and @ingve - what do you think?

ingve:
Have you tried other GSL implementations?

GSL Lite works with GCC 4.9.2: 

https://github.com/martinmoene/gsl-lite

Mattia Basaglia's also has an implementation:

https://github.com/mbasaglia/Cxx-MiscLib/blob/master/include/misclib/gsl.hpp

mnordsletten:
Ok to test

PR #707, opened 2016-08-11
New test structure implemented

mnordsletten:
Added .keep file in the /unit folders to force git to sync these files as well.

The folders that don't follow the new structure are:
```
├── examples
│   └── build
├── mod
│   └── gsl
├── stress
│   └── stress
```

The rest have the agreed upon unit and integration folders present.


mnordsletten:
Ok to test

Changed Std Lib tests -> stl tests
The failed Std Lib tests status will not be updated.

alfred-bratterud:
Cool, here we go, expect KABOOM :-)

PR #709, opened 2016-08-11
Updated test files to work with new structure

mnordsletten:
Two parts have been changed:

1. Changed the path to the vmrunner.py script. It now uses the INCLUDEOS_SRC variable instead.
2. The Makefile of the projects that use lest now address lest correctly. Again I decided to use the INCLUDEOS_SRC variable here.

Let me know if number 2 is not done correctly. Not made many Makefiles.


mnordsletten:
- Stress test is failing on the arping number 7 as we have seen before @fwsGonzo 
- Util test is failing because it is not finished yet. No test.py present. I've disabled it until we actually have a working test in that folder.
- Build examples test is failing because of some error in the script. This one should be easy to fix. 
- The network test was unstable. It was stuck on the transmit test. This is the one that was not fixed yet right @alfred-bratterud ?

This one is ready for merging, as it fixes a lot of the integration test issues the new structure has made.

PR #713, opened 2016-08-11
[wip] Build system

andreashappe:
Hi, it made no sense to split between fedora-fixes and cmake branches (as I wasn't able to really compile my cmake branch due to missing fedora fixes, meh).

Okay, I'm trying to create a minimal CMake setup running. Currently it (kinda builds):

* multiboot (with lots of warning due to cmake passing the same options to linker and compiler. GCC expects this, clang not so much)
* os.a (I do not believe that it actually works)
* libc++abi.a (which IMHO also is not working)

bootloader is a PITA right now. But then, as soon as that is also installed, debugging in earnest can start.


andreashappe:
could be worse, to run it go into the src folder and do the following:

~~~ sh
$ mkdir build
$ cd build
$ CXX=`which clang++` cmake ./../
$ make
~~~

This should build `os.a` (named `libos.a`), `libc++abi.a`, `bootloader` and `multiboot`.

RicoAntonioFelix:
Hey @andreashappe ... Nice effort in adding value to the project :+1: :+1:

Question: Why is the [install directory](https://github.com/andreashappe/IncludeOS/blob/2c55987b4531199bd21072848266b174f8051d32/src/CMakeLists.txt#L5) hard-coded to your `${HOME}`? ... Probably just something that was overlooked :wink:

andreashappe:
that's the `[WIP]` part (; also I have the the stack protector value hard coded, etc.

mnordsletten:
Good work so far :) I'm testing it now on both Ubuntu 16.04 and Ubuntu 14.04. I might just take it for a spin on Mac OS X as well. 
Expect a pull request soon as I fixed a few things.

alfred-bratterud:
@andreashappe wow, nice job! @mnordsletten glad you're on it - looking forward to giving it a spin myself :-)

andreashappe:
another question regarding the build system: would it make sense to juse `jq` instead of python for extraction JSON information (unik is doing the same)?

mnordsletten:
I think adding the possibility of using JSON in more places is a very good approach. We are currently using it for our test system through python, but using `jq` in other circumstances sounds good.

@andreashappe I sent you a pull request to your fork.

alfred-bratterud:
@mnordsletten any jenkins on this PR? Tell me when to merge.

mnordsletten:
Ok to test

I'll monitor to see what happens. I'll let you know when we can merge.

andreashappe:
hm, lest.hpp?

mnordsletten:
If you merge in the latest from includeos/dev there is a fix for the missing lest.hpp. Then we can do another run of the jenkins tests.

mnordsletten:
Ok good. As we just moved to a new test structure yesterday the build examples test and the stress test are not functioning as intended. The rest of the tests are working so on my behalf this is ready for merging @alfred-bratterud

andreashappe:
please don't merge yet, the CMake stuff (while being optional) wasn't really tested at all

andreashappe:
another question: what's the relationship between ./etc/install_from_bundle.sh and ./etc/install_build_only.sh ? Because the scripts within install_build_only.sh looks a bit redundant -- and if install_build_only.sh is intended as "stand alone" script, then IMHO it should be moved into the project's root directory (or in a new scripts/ directory)

andreashappe:
@mnordsletten  regarding using jq instead of the python script:

`jq '.[0]["assets"][0]["name"]' github.json`  would extract the release-name -- think that's better readable than having a full python script to achieve that, what do you think?

mnordsletten:
The entire ./etc directory could use a spring cleaning, lots of scripts are from the early days. I'll get around to it eventually. I'm sure there are lots of conflicting scripts that do partially the same thing. Any suggestions are welcome of course :)

mnordsletten:
@andreashappe Yeah jq command does indeed look a lot smoother. Just make sure that jq is installed automatically before it is used as it is not a default package.

andreashappe:
@alfred-bratterud can you install jq with `brew install jq` on the other operating system?

mnordsletten:
@andreashappe `brew install jq` is confirmed to work on Mac OS x 10.12. I just tested it.

andreashappe:
When was ./etc/install_build_only.sh last checked? It seems like that it uses the python script wrong (ASSET might always be NULL). Also it was only run on ubuntu (due to the `apt-get` usage).

@mnordsletten : seems like the automatic install of jq did fail on the jenkins.

alfred-bratterud:
@AndreasAakesson https://github.com/hioa-cs/IncludeOS/blame/master/etc/install_build_only.sh: I think this is a part of the old docker based build system. @ingve has replaced that with a new one yes?

ingve:
Yes, we have a separate Docker build system for our UniK compilers (piggy-backing on the default install.sh script).

mnordsletten:
The build examples test is now working correctly in a different pull request I have submitted. The rest looks good :)

andreashappe:
@ingve so I could remove that script?

ingve:
@andreashappe It is not needed for our Docker build workflow, but it could be used/referenced from other places. @mnordsletten is the final arbiter for any file that ends with .sh

mnordsletten:
You can go ahead and remove it. At this moment removing scripts that clutter and cause confusion is a good way to go.

andreashappe:
other candidates for "clean-ups":

* apply_license.sh
* license_header.txt
* license_header_old.txt
* rename_tag.sh

move to another repository and delete? Not used by the project.

andreashappe:
okay the meeting is closing down.. expect less commits from me (;

alfred-bratterud:
@mnordsletten is this OK to merge from your side? @ingve - docs from my email in the scripts - you approve? Otherwise there's an FAQ on the wiki, which is easier to maintain without changing the commit hash (only important for master). But it's better to merge this now than to keep it hanging if neither of you have any objections.

andreashappe:
@ingve: stupid question: a github wiki can be checkout through gitlab. Would it be possible to just add the gitlab wiki as /docs directory through a git submodule? That way a potential user already has the documentation installed after he has run install.sh

ingve:
Having huge front-matter comments in English prose *in source files* is an antipattern. Yes, they did this back in the 60s when they wrote everything in assembler, but it is not a good idea today. People expect to find docs collected in one place; we shouldn't make them trawl through lots of source files to find documentation. Nobody will remember to update all that text when changes to the code are inevitably made, and we'll end up with comments saying one thing and the code doing another.

The right place for significant amounts of documentation is in the documentation section of our web site. I suggest we set up a new Jekyll/GitHub pages combination, seed them with the existing wiki content, and start taking pull requests for proper docs.

andreashappe:
@ingve so I guess you're not a fan of literate programming (;

I would expect documentation to be "at hand" -- which in this case was as I did not understand why all the libraries had to be downloaded instead of just using the local system's libraries. Also I do not see why having separated documentation and code improves the documentation's freshness -- I have rather seen the opposite in software projects.

I bow to the project's preferences, but could you please add a short note with a link to the documentation to the download script? Without any documentation it's a bit tough to decipher why all the stuff is going on.

andreashappe:
@ingve : to be honest I don't understand why this is an anti-pattern? Documentation that describes what the code does should be removed (for that you have code), but I have never seen that documentation that describes why code is here in the first place was seen as code smell or cruft. Do you have a reference for that, I am seriously (and honestly) interested

ingve:
As mentioned up-thread, @mnordsletten is the final arbiter for any file that ends with .sh. I defer to his judgement.

alfred-bratterud:
Hi @andreashappe I'm looking at this now. I don't have any strong opinions on the documentation issue - having it in code is really nice some times. The only issue for me is that a change in documentation will then have to be a new commit, which again means that the release tag will have to be changed in order to have people pulling the "latest release" get the updated docs. I'm sitting down to test your build now and if it works I'll merge. Again, thanks a lot for the work you've done on this!

alfred-bratterud:
Hi @andreashappe unfortunately the `libos.a` doesn't work. When I'm trying to link against it I'm getting things like 
```
d: warning: cannot find entry symbol _start; defaulting to 0000000000200084
/home/alfred/IncludeOS_install/multiboot.o:(.rodata+0x1c): undefined reference to `_start'
/home/alfred/IncludeOS_install/os.a(syscalls.cpp.o): In function `gettimeofday':
/home/alfred/IncludeOS/src/kernel/syscalls.cpp:(.text+0x3ae): undefined reference to `RTC::get()'
/home/alfred/IncludeOS_install/os.a(syscalls.cpp.o): In function `clock_gettime':
/home/alfred/IncludeOS/src/kernel/syscalls.cpp:(.text+0x4c3): undefined reference to `RTC::get()'
/home/alfred/IncludeOS_install/os.a(os.cpp.o): In function `OS::start(unsigned int, unsigned int)':
```
indicating that it doesn't contain all the objects required. Any ideas? Let me know how you want to proceed with this.

fwsGonzo:
ok to test

andreashappe:
hi, I rebased the commits into https://github.com/hioa-cs/IncludeOS/pull/734

I squashed some bug fixes and reverts (for newly introduced stuff) and kept the cmake/doc stuff out of it. I can should be able to rebase this pull request against that later on (as mostly CMakeLists.txt files are touched).

andreashappe:
@alfred-bratterud : is this getting better? One problem was using nasm with `-f bin` and `-f elf`.

fwsGonzo:
ok to test

andreashappe:
@mnordsletten : something I encountered during reading the code:

~~~
STACK_PROTECTOR_VALUE=$(shell awk 'BEGIN{print int(rand()*65536)}')
~~~

will always produce the same value (deterministic random number generator). You might want to call `srand()` to reinitialize the RNG beforehands. I have added that to the cmake-branch.

Regarding the branch: by now I can use it to build the demo_service image (with just a small amount of cheating and multiple ugly hacks). Question is: if, how and when should this be integrated into the dev-branch? So far it's totally separated from the Makefile-based system so both build systems can be used in parallel.

fwsGonzo:
Yes, I know about it and I wish there was a better easier way, but its not terribly important that its "completely" random. The thing about is that only we rebuild the OS, and it happens semi-rarely, and on top of that, the value itself is only used in the `crt/c_abi.c` file, where the stack protector value is part of read-only data segment. After that, each function marked for stack-protection gets that value as its initial value. So, all in all, the value doesn't have to be that much random, it just needs to vary from second to second, so its possible to use some modified timestamp too. The only real improvements we want to do with the stack protector is to make the random value work on more systems, and to randomize the `memory location` of the stack protector value in read-only section. Not sure how to do the last one.

fwsGonzo:
This is the output from the `test` script that builds our examples folder:
```
+ cd /var/jenkins/workspace/pull_request_build/test/examples/build
+ ./test.sh
>>> Will now attempt to make all examples. Outpt from make will only be present if an error occured


>>> Now making build
make: *** No targets specified and no makefile found.  Stop.
```
Not sure why that's happening

andreashappe:
@fwsGonzo : well, when using gawk (which should be the default on linux) the value is not somewhat random, but fixed. If you call srand() before rand() then you get the time-based behaviour (changing every second, i think this is what mawk does as default)

andreashappe:
@fwsGonzo : hm, within the test.sh script, line 30: it should be $script_absolute_dir/../../../examples instead of $script_absolute_dir/../../examples. But I am not sure why its behavior (or execution path) should have changed. All the CMake changes will not used during the test run (as it uses the Makefile based build system)

fwsGonzo:
The examples build script works for me here, as is. So I'm not sure what's happening..

andreashappe:
@fwsGonzo : same error happens on my (Fedora Linux) system, strange.

I wanted to add support for compiling some missing stuff (src/crt is currently taking from ~/IncludeOS_Install) and then create a new fork containing a single commit with all the new CMakeFiles. At least all unikernels (sans the STREAM one) can be built with cmake (on my machine), so that might make it a good starting point

fwsGonzo:
This project is "fairly" advanced: https://github.com/fwsGonzo/IRCd
It's a basic IRC service, so it uses more facilities than normal example services

alfred-bratterud:
@andreashappe if the stuff in here has been split out to separate PR's I guess this PR is now due for closing?

andreashappe:
yes. will close

PR #726, opened 2016-08-16
intr: Store eflags

fwsGonzo:


zwparchman:
If I'm reading the [intel manual](http://www.intel.com/content/www/us/en/architecture-and-technology/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.html) (iret starts at page 603) correctly it looks like iret already restores eflags. Other sources suggest the same. Am I mistaken?

fwsGonzo:
sounds reasonable, will undo. thanks for the tip :+1:

PR #730, opened 2016-08-17
test: verify no clear bits can be found when all bits are set

ingve:
Increases test coverage for membitmap.hpp from 96.4% to 98.2%


fwsGonzo:
The tests are passing now, I'll just merge this

Issue #718, opened 2016-08-12
Universal access point for IP stacks


alfred-bratterud:
Currently the user instantiates IP stacks directly on top of devices. This works fine but not if you have separate modules that needs access to the same IP stack - then you'd be replacing the delegates for that device to point to your new stack and essentially stealing traffic from the other. I suggest a template factory method in `Inet` that provides static instances of say `Inet<DRIVER>(Nic<I, DRIVER>)`. The factory should be the only allowable way to create Inet instances.

alfred-bratterud:
@AndreasAakesson close?

AndreasAakesson:
Since we only got IPv4 stack atm, I guess it's okay. What do you think?
It's not all the way down to `net::Inet`, but `net::Inet4`

alfred-bratterud:
That's for 2017 :-)

PR #740, opened 2016-08-19
Driver decoupling

AndreasAakesson:
**Inet4 is no longer a template class**
Took a lot of rewriting, but turned out good.

**hw**
* Renamed `Dev` to `Devices`
 * Global access point for all *registered* devices (devices initialized with a driver).
 * `Devices::nic(N)` and `Devices::drive(N)`
* Renamed `IDiskDevice` to `Drive`
* `Nic` is now base class for all Network drivers (currently VirtioNet)
 * Used by the user - now longer needs to know about the driver
* `Drive` is now base class for all Block device (Drive) drivers (currently VirtioBlk)
 * Used by the user - now longer needs to know about the driver

**net**
* Universal access point for Inet4 stacks (solves #718?)
 * `Inet4::stack<N>()` - constructed when called for the first time. Only way to bring up an IPv4 stack now.

**kernel**
* Introduced *drivers* - compiled separately from the kernel.
 * `VirtioNet` and `VirtioBlk` now using global constructor to register its driver.
 *  Which driver to include in service is now defined in Makefile: `DRIVERS=virtionet`
* `PCI_manager` now stores driver "factories".
 * If a driver factory is found when probing the PCI bus, register the device at `hw::Devices`

**Other**
* `Service::start` can now make us of commandline parameter
* Automatically deduce service image name c3ce196
* Experimental randomization of addresses (?)


AndreasAakesson:
ok to test

AndreasAakesson:
`test/fs/integration/memdisk/bigdisk` now fails on a VM with 256MB memory and a memdisk of 256 000 sectors (~128MB):
```
[ FAIL ] <VMRunner> terminating with uncaught exception of type Memory_range_exception: Maximum range size is 2147483647
```

PR #743, opened 2016-08-20
Delegates: added explicit null-initialization 

alfred-bratterud:
Trying to make it behave more like std::function. One difference is that I've kept our default initialization to be an empty delegate. This might be the wrong choice - it will return a default constructed object of whatever the return type was.  This was done because people (including myself) notoriously forgot to initialize delegates. Now that we throw if you try to call a null-initialized one that might be the better solution. Any opinions?


AndreasAakesson:
I approve this suggestion :+1:

alfred-bratterud:
I'll clean up and do this in a separate PR, and also remove default initializatoin of delegates to empty lambda.

PR #746, opened 2016-08-22
Memory map update (including test) + delegates default to nullptr and throws if then called

alfred-bratterud:
…empty lambda initialization


alfred-bratterud:
test please

PR #752, opened 2016-08-23
[build] Build libos.a as and unikernels within /example with cmake

andreashappe:
This is a rough first prototype that allows building of unikernels
within examples/ with cmake. Lots of room for improvements. Also
an installed ~/IncludeOS_Installation is expected.

To build (on Linux), goto to the unikernel src directory, and do:

~~~
 $ mkdir build
 $ cd build
 $ CXX=/usr/bin/clang++ CC=/usr/bin/clang cmake ./..
 $ make
~~~

To switch configuration switches, use ccmake within the build directory.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -17,14 +17,15 @@ add_custom_command(
 
 # TODO: i wanted to use a glob, but then found out that not everything is included
 set(OS_OBJECTS  kernel/start.asm kernel/kernel_start.cpp kernel/syscalls.cpp
-		kernel/interrupts.asm kernel/os.cpp kernel/cpuid.cpp kernel/memmap.cpp
-		kernel/irq_manager.cpp kernel/pci_manager.cpp
+		kernel/interrupts.asm kernel/os.cpp kernel/os_stdout.cpp kernel/cpuid.cpp
+		kernel/memmap.cpp kernel/irq_manager.cpp kernel/pci_manager.cpp
 		kernel/elf.cpp kernel/terminal.cpp kernel/terminal_disk.cpp
-		kernel/vga.cpp kernel/debug_new.cpp kernel/profile.cpp
-		kernel/timer.cpp kernel/rtc.cpp util/memstream.c util/async.cpp
+		kernel/vga.cpp kernel/debug_new.cpp
+		kernel/profile.cpp kernel/profile_intr.asm
+		kernel/timers.cpp kernel/rtc.cpp util/memstream.c util/async.cpp util/statman.cpp
 		crt/c_abi.c crt/string.c crt/quick_exit.cpp crt/cxx_abi.cpp crt/mman.cpp
 		hw/ide.cpp hw/pit.cpp hw/pic.cpp hw/pci_device.cpp hw/cpu_freq_sampling.cpp
-		hw/ps2.cpp hw/serial.cpp hw/cmos.cpp
+		hw/ps2.cpp hw/serial.cpp hw/cmos.cpp hw/drive.cpp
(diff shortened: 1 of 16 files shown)


alfred-bratterud:
test please

alfred-bratterud:
@andreashappe great - I just merged my own PR which should fix the storage tests in jenkins. There are still issues with its internal networking / bridge it seems, but at least now we should get further

alfred-bratterud:
test please

alfred-bratterud:
@andreashappe / @mnordsletten we've fixed all the tests now, but this still fails. Any ideas?

mnordsletten:
Yeah, this is because a CmakeList.txt file was added to the examples directory. The build examples test looks for all directories, but there is a mistake in the line that looks for directories. 
Add an extra / at the end of line 29 in IncludeOS/test/examples/build/test.sh and it should work.
```
for dir in `ls -d $script_absolute_dir/../../../examples/*/`
```

alfred-bratterud:
@mnordsletten / @ingve ready to merge?

andreashappe:
Hi, how are we doing with this pull request? Is there an upcoming release after that we could merge this stuff?

alfred-bratterud:
test please

andreashappe:
okay, so this is how a `git rebase` that did went wrong looks like.

andreashappe:
@AndreasAakesson : I believe `__BEGIN_DECLS` is defined in the (removed) feature.h. I've added the following code to fix compilation (at least so far):

~~~
/* C++ needs to know that types and declarations are C, not C++.  */
#ifdef __cplusplus
# define __BEGIN_DECLS extern "C" {
# define __END_DECLS }
#else
# define __BEGIN_DECLS
# define __END_DECLS
#endif
~~~

src/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+
+add_custom_command(
+	OUTPUT apic_boot.o
+	COMMAND nasm -f bin -o apic_boot.bin ${CMAKE_CURRENT_SOURCE_DIR}/hw/apic_boot.asm
+	COMMAND objcopy -I binary -O elf32-i386 -B i386 apic_boot.bin apic_boot.o
+	DEPENDS hw/apic_boot.asm
+)
+
+# TODO: i wanted to use a glob, but then found out that not everything is included
+set(OS_OBJECTS  kernel/start.asm kernel/kernel_start.cpp kernel/syscalls.cpp
```
RostakaGmfun: How about splitting these files into CMake object libraries by folder?
E.g. `add_library(kernel OBJECT kernel/interrupts.asm...)`
This will improve readability a lot and will not impact the build process anyhow.

andreashappe: (sorry for the late reply, I've been hopping between conferences)

sounds good! so far I'm mostly in a "trying get a minimal CMake conversion up-and-running" mode --- the missing step is adding more validations that the cmake build system creates actually working images.

I will look into this, as soon as the initial version is merged. If you want to, you can add a patch doing that to the pull request (and I would include that change to my pull request).

andreashappe:
hm. what's the next steps before this can get merged (or can we merge even before)? As the build system touches everything it gets out-of-sync quite fast..

The big things IMHO currently missing:

* installation of library into ~/IncludeOS
* support for external projects (e.g., the IRC server)

Also there's lots of small stuff that needs to be done (clang selection, compile flags, etc.) but those can be done incrementally. I think it would be best to get this merged and, even more perfect, if one or two core developers would start using cmake (to get some ongoing testing done).

Also, last time i used cmake was approx. 7 years ago --- so I expect if someone else would look into the cmake stuff, there might be cleaner ways of solving stuff. But this should be a good initial starting point.

PR #753, opened 2016-08-23
Statman: Added Statman functionality

AnnikaH:


api/utility/statman.hpp
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    return stats_.end();
+  }
+
+  auto begin() { return stats_.begin(); }
+
+  auto end() { return stats_.end(); }
+
+  auto cbegin() { return stats_.cbegin(); }
```
fwsGonzo: I believe the const iterator functions should be marked const:
```
auto cbegin() const { ... }
```

PR #771, opened 2016-08-29
Nic/VirtioNet and Statman

AnnikaH:
Changed initialization of stats to VirtioNet constructor. Stat getters returning reference instead of pointer. Statman unit test updated.


alfred-bratterud:
Nice!

PR #781, opened 2016-08-30
IP4: Now dropping packets with destination address not matching local…

AnnikaH:
… address. Stat for packets dropped added.


alfred-bratterud:
Great @AnnikaH, nice find 👍

PR #799, opened 2016-09-01
Allow memory size above 2GB! 

alfred-bratterud:
The heap can't use this atm. due to ptrdiff_t being signed, but that's just temporary (memory map still WIP). But I couldn't live with vm's not being able to boot with more memory, so now they can. The memory above heap can then be used freely by the service.


test/vmrunner.py
```diff
@@ -253,30 +253,33 @@ def __init__(self, config, hyper = qemu):
...
+    self._on_output["SUCCESS"] = callback
 
   def on_panic(self, callback):
-    self._on_panic = callback
+    self._on_output["PANIC"] = callback
 
   def on_timeout(self, callback):
     self._on_timeout = callback
 
-  def on_exit(self, callback):
```
AndreasAakesson: Accidentally overwritten?

PR #807, opened 2016-09-02
Fixed DHCP replies being filtered out

alfred-bratterud:


alfred-bratterud:
I've tested this on: OS X running in Virtualbox, confirming that you get DHCP. All integration tests and all unit tests pass on Ubuntu 16.04.

PR #818, opened 2016-09-07
clang++ as compiler binary name is supported for seed projects

spjoe:
In src/Makefile clang++ is already allowed as compiler, but not in seed projects.

This change would be nice for arch linux users.


alfred-bratterud:
Hi @spjoe thanks! I'll get this merged with the next update - we just found and solved some more issues

PR #829, opened 2016-09-13
Changes to filtering of test categories

mnordsletten:
test.py feels like a bit of a mess. Integrating the test categories into the test class will probably be a better option. Also integrating both the validate_test and validate_all script into at least the same script will be a good move. As of right now this is hard to maintain. 

This makes changes to how jenkins calls tests, but will require less manual interaction with Jenkins in the future.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -228,7 +228,7 @@ def main():
     test_folders = []
   tests_combined = test_categories + test_folders
   if args.tests:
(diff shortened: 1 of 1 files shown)


mnordsletten:
Ok to test

mnordsletten:
Ok to test

mnordsletten:
Ok to test

mnordsletten:
Ok to test

mnordsletten:
Ok to test

PR #833, opened 2016-09-21
Update Readme with link to Acorn

ingve:


alfred-bratterud:
Hi @ingve thanks for the PR - we were way too busy to do this during the conference, so can you change this to a more general notice?

PR #836, opened 2016-09-30
Added Logger utility + build stuff

AndreasAakesson:
* `seed/` folder is now being installed (IncludeOS_install)
* Service makefile can now add custom libraries: `EXTRA_LIBS=`

Logger works but there are improvements to be made. Reading entries is a bit expensive atm.


mnordsletten:
I'm looking into the failing dhcp test. I've seen this a few times now, but never been able to reproduce. Ok to test

PR #840, opened 2016-10-03
Add 256 color vga demo

lefticus:
Proof of concept for setting VGA 256 color mode.

1. Adds ability to specify nasm files for assembling and linking into project
2. Adds example that jumps into 16 bit mode to call interrupt 0x10 to set VGA mode and returns back to 32 bit mode and continues execution
3. Draws mandelbrot set to video memory

The VM resets repeatedly if the network driver is enabled. I don't know if the VGA address space of 0xa0000 somehow conflicts with the rest of includeos?

### DIFF
--- a/etc/install_from_bundle.sh
+++ b/etc/install_from_bundle.sh
@@ -47,7 +47,7 @@ popd
 # Build IncludeOS
 echo -e "\n\n>>> Building IncludeOS"
 pushd $INCLUDEOS_SRC/src
-make -j
+make -j2
 make install
 popd
 
--- a/examples/256_color_vga/Makefile
+++ b/examples/256_color_vga/Makefile
@@ -1,0 +1,27 @@
+#################################################
+#          IncludeOS SERVICE makefile           #
+#################################################
+
+# The name of your service
+SERVICE = 256_Color_VGA
+SERVICE_NAME = 256 Color VGA
+
+# Your service parts
+FILES = service.cpp
+ASM_FILES = int32.asm
+
+# Your disk image
+DISK=
+
+DRIVERS=
+
+# Your own include-path
+LOCAL_INCLUDES=
+
+# IncludeOS location
+ifndef INCLUDEOS_INSTALL
+INCLUDEOS_INSTALL=$(HOME)/IncludeOS_install
+endif
+
+# Include the installed seed makefile
+include $(INCLUDEOS_INSTALL)/Makeseed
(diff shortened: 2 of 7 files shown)


fwsGonzo:
Always glad to see some graphics programming making its way into here. :+1: 

Looking at the makefile made me realize the way we do it now we can't compile with C-compiler or assembler, and there is no need to add more lines to makefile when we can use extensions to determine the program to use. :)

For example the assembly file would be listed as int32.o which makes the makefile try to complete a recipe, and it will find that int32.asm can be turned into int32.o.

As for the other problem:
0xa000 (4 less bits) is used by ELF symbols. Adding the network driver (and some networking code along with it) will instantiate alot of code and symbols. It's clear that color mode VGA will get overwritten here, unfortunately. We can fix it by moving ELF symbols on the heap. The only reason we kept them at 0xa000 was because we could :) Trying to use some of that reserved 1mb area for something good.
In crt/c_abi.c the magic location is simply set. There are a few hidden functions in kernel/elf.cpp with C calling convention that can be used to malloc and move the symbols into heap before calling global constructors.

Also, we would like to apply new changes to the dev branch, which is where we work until new releases are finalized. Does github support rerouting PRs to other branches now?

alfred-bratterud:
Awesome @lefticus, great to see some pixels and fractals :-) The switch to 16-bit mode for BIOS calls is also nice - probably what we should do to get the memory map correctly populated when not using multiboot. (With multiboot we get this info prefetched from the bootloader). I've merged this into dev-branch now.

lefticus:
@fwsGonzo I tried many different ways to make the recipe smart enough to decide how to compile asm/C++ files on its own without having to add a separate ASM_FILES variable. I just simply could not get it to work (I'm not well versed in Makefiles, I have not used them in ~7 years since moving to CMake).

But I think that would be the better decision. The thing that I was a little lost on also was link order. I know gcc/clang can be sensitive to that, and didn't know how that would work if you had all of the files in the same variable.

I could have moved the PR to `dev`, but since you already merged the commits there github will not let me now "there are no differences between your branch and dev" so we can just close this PR.

Also, I did strongly consider adding the jumps in/out of 16-bit mode into your kernel, but it's not code I own. It is CC licensed, however.

So, just be aware, if you go to use this code more globally to make sure the license is compatible with what you have:

https://github.com/hioa-cs/IncludeOS/blob/dev/examples/256_color_vga/int32.asm#L2-L11

Otherwise, I would think it is safe in the examples folder, for how I understand software licensing.

fwsGonzo:
Hey! Yes, it's all merged now, and I also made it possible to just use "int32.o" as a sort of recipe target that automatically builds the assembly file without needing ASM_FILES.
Closing this now
Now to figure out what happened when network driver is added. :)

PR #844, opened 2016-10-05
Fix bug to allow reading large (>4kish) files.

danieloneill:
Simply initialise VirtioBlk::inflight to 0 in the constructor, to allow reference count to reach 0 (and fire the callback).


fwsGonzo:
Nice catch :+1:

alfred-bratterud:
ok to test

Issue #620, opened 2016-06-28
Make install script more robust


ingve:
Currently, the install script spews a lot of output as it is downloading and building IncludeOS, and no matter what goes wrong, the script valiantly trudges on. If there is a networking issue during the download phase, the script will still run to completion and claim that installation is done, even if it was unable to build os.a and copy it to the destination directory. With hundreds of lines of output to scroll through, it can be difficult to identify the problem.

Would it be possible to have some basic sanity checking in the install script? After installing dependencies, check that they can in fact be run? After building os.a and friends, check that they actually exist before outputting the "installation done" message, etc.

fwsGonzo:
We could download the checksum files also and validate all the downloads, although it would make the install process a little bit longer.

andreashappe:
The cmake-based download+build has support for checksum verification (mostly, so that fiels will not be downloaded again and again).

But I've only tested it with Fedora (Linux), no virtual network device is created (yet) and some stuff is missing. But it might be fixable before the next release.

HTH

PR #848, opened 2016-10-05
Adds missing exit value checks in install script

gurka:
So this is basically just some additional return / exit value checks in the install scripts to make it more robust. I found most of these problems when trying to install IncludeOS under Debian stable. Most of the build steps failed but the script kept going, and finally outputted "Done! Test your installation with ./test.sh", which was kind of .. not so good. This commit partially solves #620 .

I removed the "sudo" part from the README since the script will use sudo when needed. Using sudo when executing install.sh will also build and install everything as root:root and that is probably not want most users want.

I'm not sure if the check for "sudo" is necessary since all supported systems have sudo installer per default (?). Debian does however not, and that's why I added it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,10 +42,12 @@ A longer list of features and limitations is on the [wiki feature list](https://
 
 ### Install libraries
 
+**NOTE:** The script will install packages and create a network bridge, and thus will ask for sudo access.
+
 ```
     $ git clone https://github.com/hioa-cs/IncludeOS
     $ cd IncludeOS
-    $ sudo ./install.sh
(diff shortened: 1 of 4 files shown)


AndreasAakesson:
Yeah, sudo should not be used with install.sh - because of the reason you state; everything gets build and installed as root. Thanks for the PR :)

ok to test

AndreasAakesson:
retest please

mnordsletten:
This looks great. Having it exit at the point of failure will make it a lot easier to identify the error.

PR #850, opened 2016-10-06
test.py improvements

mnordsletten:
A little rewrite of the test.py script that basically adds two distinctions:
- `test_categories`: fs, net, kernel etc.
- `test_types`: integration, stress etc.

I've also added the skip reason to the Test class so it will now be associated with a test object. Allowing for easier management when printing the reason for skipping. 

This still does not allow the running of individual tests.


mnordsletten:
Ok to test

mnordsletten:
Ok to test

PR #852, opened 2016-10-06
readme: updated logo, added links to Acorn and Mana

ingve:


mnordsletten:
All these random dhcp test fails should be sorted out with the newest updates to dev.

We can try and run the tests again here, ok to test. If the issue persists, just pull the latest from dev and it should get sorted.

PR #857, opened 2016-10-07
net: Work on isolation/layering - Ethernet & IP4

AndreasAakesson:
Just a start.

* Ethernet address => hardware address `hw::MAC_addr`
* Extracted IP4 header

Partly removed Ethernet dependency on Nic, and IP4 dependency on Packet


AndreasAakesson:
retest please

PR #859, opened 2016-10-07
Snake example rewrite

Voultapher:
My design goals were:
Clean modern C++
Clean task separation
Better user experience
Better readability
No vga read
No raw loops
Little control flow change with if
No usage of rand()

### DIFF
--- a/examples/snake/service.cpp
+++ b/examples/snake/service.cpp
@@ -16,31 +16,27 @@
 // limitations under the License.
 
 #include <os>
-#include <vga>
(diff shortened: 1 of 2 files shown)


fwsGonzo:
Looks good! The `hw::PIT::on_timeout_d` class is deprecated and will in the future just be more hardware oriented in its API, but we already have a `Timers` class which has microsecond precision. If you could use that and then forward any PRs to our dev branch, that would be nice :) :+1:

Voultapher:
Ok will do.

PR #861, opened 2016-10-08
kernel: draft of cpuid rewrite

gurka:
First draft of cpuid rewrite. Open questions:

- Is the cache necessary? I did some research and found that cpuid is a pretty expensive call, but maybe the overhead of the cache makes it useless.
- Maybe add vendor information to each `FeatureInfo` since not all features are applicable to both Intel and AMD CPUs. At the moment the user needs to first check `isIntelCpu` or `isAmdCpu` before asking for a specific feature (if the feature is not applicable to both).

### DIFF
--- a/api/kernel/cpuid.hpp
+++ b/api/kernel/cpuid.hpp
@@ -19,10 +19,86 @@
 #ifndef KERNEL_CPUID_HPP
 #define KERNEL_CPUID_HPP
 
-struct CPUID {
-  static bool isAmdCpu();
-  static bool isIntelCpu();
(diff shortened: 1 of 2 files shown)


fwsGonzo:
Looks really nice :) Now we just need the kvm paravirtual stuff, from here: https://www.kernel.org/doc/Documentation/virtual/kvm/cpuid.txt

I tried making PV_EOI work, but it's not setting the bit for me. I suspect we have to check the feature first maybe. Or maybe it's just not enabled

fwsGonzo:
ok to test

PR #862, opened 2016-10-08
Snake example rewrite

Voultapher:
My design goals were:
Clean modern C++
Clean task separation
Better user experience
Better readability
No vga read
No raw loops
Little control flow change with if
No usage of rand()


fwsGonzo:
ok to test

PR #867, opened 2016-10-11
Add support for Arch Linux (and its flavors).

GuzTech:
This simple patch enables setting up IncludeOS on Arch Linux and its flavors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,14 @@ A longer list of features and limitations is on the [wiki feature list](https://
     $ sudo ./install.sh
 ```
 
+Or if you are running Arch Linux (or one of its flavors):
+
+```
+    $ git clone https://github.com/hioa-cs/IncludeOS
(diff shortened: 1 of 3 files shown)


gurka:
It looks good! I don't think there is a need for extra install information in README.md since it's the same steps as the other platforms (on dev branch there is no sudo before ./install.sh).

GuzTech:
The only reason I added the extra install information is because I first did `sudo ./install.sh` but that installed IncludeOS in `/root/IncludeOS_install`, and `./test.sh` expects that folder to be in the home directory of the currect user, and not that of root. The install script already does use sudo whenever it needs to, so there's no need to call it with sudo.

Of course you can change the patch in any way you like :)

fwsGonzo:
I merged this into dev, which will be merged into master soon-ish :) Thanks for the fix :+1:

PR #871, opened 2016-10-13
Changed Packet to be sent as unique_ptr instead of shared_ptr

AndreasAakesson:
Also; TCP Connection is now a tiny bit smaller, and no longer allocates default callbacks.

### DIFF
--- a/api/kernel/cpuid.hpp
+++ b/api/kernel/cpuid.hpp
@@ -21,7 +21,7 @@
 
 namespace CPUID
 {
-  enum Feature
+  enum class Feature
   {
     // ------------------------------------------------------------------------
     // Processor Info and Feature Bits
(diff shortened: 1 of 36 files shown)


api/kernel/cpuid.hpp
```diff
@@ -21,7 +21,7 @@
 
 namespace CPUID
 {
-  enum class Feature
```
gurka: Why remove scoped enum?

fwsGonzo: I think it was because I couldnt convert it to an int, or something

gurka: Aha, yeah you have to` static_cast<int>` it if it's an enum class. (or `static_cast<T>` if it's underlying type is explicitly set to `T`, e.g. `enum class Foo : uint8_t {};` )

PR #873, opened 2016-10-13
install: do not use sudo if not needed

gurka:
* Adds check of installed packages before calling sudo apt-get
  (Ubuntu only)
* Adds check of created / configured bridge before calling
  sudo brctl and sudo ifconfig

You can now run ./install.sh without needing to enter password for sudo if everything is already installed.

**NOTE:** Can someone test this on other platforms? I'm not 100% sure if OS X, Fedora and Arch has the command `ip`. Otherwise we need to add a check for that and always configure the bridge, even if it's not actually needed.

### DIFF
--- a/etc/create_bridge.sh
+++ b/etc/create_bridge.sh
@@ -1,23 +1,43 @@
-#! /bin/sh
+#!/bin/sh
 
 BRIDGE=include0
 NETMASK=255.255.0.0
 GATEWAY=10.0.0.1
 
+# Håreks cool hack:
+# - First two bytes is fixed to "c001" because it's cool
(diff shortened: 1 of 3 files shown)


gurka:
Maybe it doesn't matter if `ip` is available. If it's not the expression will always evaluate to false and the bridge will always be configured.

```
>>> Network bridge already created
./etc/create_bridge.sh: 25: ./etc/create_bridge.sh: ip_asdasd: not found
>>> Configuring network bridge (requires sudo):
-e
```

alfred-bratterud:
add to whitelist

Issue #874, opened 2016-10-13
Building services using newest master does not work out of the box on OS X 10.11


ingve:
After a fresh IncludeOS install, elf_syms is not present, and the Makefile commands to use objcopy/strip do not work under OS X without a few modifications. 

Workaround:

(0. Delete old IncludeOS, IncludeOS_install, IncludeOS_build folders)
1. Clone IncludeOS from GitHub
2. Run `install.sh`: Success!
3. Export LD_INC as instructed by `install.sh`
4. Go to examples/256_color_vga
5. Run make. Get error message "elf_syms: No such file or directory"
6. Go to vmbuilder directory, make some changes to get elf_syms to compile under OS X, make, make install
7. Go back to examples/256_color_vga
8. export OBJCOPY_INC=gobjcopy
9. Go to IncludeOS_install, make some changes to be able to use non-Apple strip
10. export STRIP=gstrip
11. make

Result: working image that can be converted by qemu-img and run under VMWare Fusion.

Changes needed to get elf_syms built:

```
diff --git a/vmbuild/elf.h b/vmbuild/elf.h
index a2d9601..cc24314 100644
--- a/vmbuild/elf.h
+++ b/vmbuild/elf.h
@@ -28,6 +28,8 @@
 #ifndef _ELF_H
 #define        _ELF_H 1
 
+#include <sys/cdefs.h>
+
 __BEGIN_DECLS
 
 /* Standard ELF types.  */
diff --git a/vmbuild/elf_binary.hpp b/vmbuild/elf_binary.hpp
index fba35cf..df3e6ef 100644
--- a/vmbuild/elf_binary.hpp
+++ b/vmbuild/elf_binary.hpp
@@ -18,7 +18,7 @@
 #ifndef ELF_BINARY_HPP
 #define ELF_BINARY_HPP
 
-#include <elf.h>
+#include "elf.h"
 #include <gsl/gsl>
 #include <stdexcept>
 ```

Changes needed to Makeseed:
```
ifndef STRIP
	STRIP=strip
endif
```
(...)
```
ifneq ($(STRIPPED),)
	$(STRIP) $(STRIPPED) --remove-section=.comment $(SERVICE)
endif
```

AndreasAakesson:
Nice detective work 👍  PR!

PR #877, opened 2016-10-14
Fixed building on macOS (master)

AndreasAakesson:
* Fixed vmbuild compile errors as explained in #874.
* Updated binutils (2.25 => 2.27) for `objcopy --update-section`
* Also installs `strip` on macOS, now set with `STRIP_INC`

Solves #874 (would be nice if someone else could confirm this)


alfred-bratterud:
Works on my mac, thanks @AndreasAakesson !

PR #880, opened 2016-10-14
(master) Update os.hpp/.cpp for includeos/acorn to run on master

AnnikaH:
Note: master branch.
os.hpp/.cpp: get_cycles_total() and get_cycles_halt() update.
Needed for dashboard component cpusage to get its data.
Please review.


alfred-bratterud:
I guess we don't want to do arithmetic on unsigned values unless we have to, but as I understand this to be necessary for acorn to work we can fix that in statman later.

PR #885, opened 2016-10-17
OS::block() + tests + optional use of int main(...) etc.

alfred-bratterud:
Just a generalization of what @fwsGonzo has been doing. Moved to separate file for use-only inclusion, added a couple of stat counters  and a couple of integration tests. 

Also some minor changes to context.


alfred-bratterud:
I'm not able to run tests locally atm. and this is pretty low-level so expect much disaster

alfred-bratterud:
@mnordsletten I don't see why Jenkins can't find `test_block.img`. It works on my end. Can you look into it?

alfred-bratterud:
..btw, I have tests for `main(argc, argv)` / `main()` but must go home

mnordsletten:
Seems like the problem is that you haven't called `vm.make()` in your test.py scripts. Without making the project manually the .img is never created.

alfred-bratterud:
Haha, ok, thanks!

mnordsletten:
Could the cause of this failing be that we are now running 5 tests in parallell? It seems like the block test is sensitive to timings. Seems like we are going to need an option that makes sure that certain tests are never run in parallell.

alfred-bratterud:
@fwsGonzo, @mnordsletten  the block test fails due to timers not having triggered enough times.  Ideas?

fwsGonzo:
It's very possible that Martin is right here, since the timers are based on deadlines (which again makes them work on any architecture). Unless the OS goes into a deep sleep and never wakes, I'm not worried.

alfred-bratterud:
@mnordsletten no space left on device?

alfred-bratterud:
Test please

PR #888, opened 2016-10-17
profile: Adds ScopedProfiler

gurka:
Notes:

* Calls cpuid before rdtsc due to http://stackoverflow.com/a/2918125/969365 (not sure if it's actually needed though)
* Not sure if we really need to add an example service. If you don't want it I can remove it.

I also did some testing before and after the "ptrstuff" commits (on bd60272f4d12c5530ef175ea472de8eca07f97fc and on adcf38c3f3df5cddd4e7592a719950df5d5fa70d):

Before:
```
--------------------------------------------------------------------------------
         29.366431 |    6013 | VirtioNet::msix_recv_handler()
         15.438494 |    3003 | VirtioNet::msix_xmit_handler()
--------------------------------------------------------------------------------
```

After:
```
 Average Time (us) | Samples | Function Name
--------------------------------------------------------------------------------
         26.893452 |    6013 | VirtioNet::msix_recv_handler()
         15.070289 |    3003 | VirtioNet::msix_xmit_handler()
--------------------------------------------------------------------------------
```

### DIFF
--- a/api/profile
+++ b/api/profile
@@ -119,6 +119,16 @@ class ScopedProfiler
     uint64_t    cycles_average;
   };
 
+
+  // The guard to use, before calling rdtsc
+  static enum class Guard
+  {
+    NOT_SELECTED,
+    LFENCE,
+    MFENCE,
+    NOT_AVAILABLE,
+  } guard;
+
   // Use plain array for performance (cache locality)
   // Increase size if there is a need to profile more than 64 scopes
   static std::array<Entry, 64> entries;
--- a/api/profile
+++ b/api/profile
@@ -7,9 +7,9 @@
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -21,6 +21,7 @@
 #define API_PROFILE_HEADER
 
 #include <cstdint>
+#include <array>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


rianquinn:
You should consider using RDTSCP instead if you can. Also, if your going to use RDTSC and CPUID, you need to set the register state for CPUID so that the execution time doesn't vary based on the register state. So for example, you could set EAX to 0 to ensure you always ask for "Basic CPUID Information".

gurka:
For some reason I thought that RDTSCP was only available on x86_64, but I realize now that it's probably available in x86_32 too - you just have to check that the cpu supports it. I got PANIC when I tried to use it :smile:.  I'll add a check and use RDTSCP if it's available.

I'll fix the CPUID + RDTSC code as well, since we'll need a fallback timer in case RDTSCP isn't available.

By the way, in some examples I've seen them change the order of CPUID and RDTSC when "stopping" the timer ( https://idea.popcount.org/2013-01-28-counting-cycles---rdtsc/ ). I don't really understand why. Why would you care about serializing the instruction queue *after* calling RDTSC?

gurka:
Oh - I guess that this explains the problem: http://akaros.cs.berkeley.edu/lxr/akaros/kern/arch/x86/rdtsc_test.c

```
 * The idea behind this was that the traditional style of using rdtsc was to
 * call:
 * 			cpuid;
 * 			rdtsc;
 * since rdtsc does no serialization (meaning later instructions can get
 * executed before it, or vice versa).  While this first cpuid isn't a big deal,
 * doing this in pairs means reading the end time also measures cpuid.  This is
 * a problem since cpuid can vary quite a bit.

...

 * Its not clear from what anyone writes as to whether or not you need to
 * serialize below rdtsc.  Supposedly, you'd need cpuid/lfence on both sides of
 * rdtsc to prevent reordering in both directions.  Andi Kleen does this in a
 * few places
```

I'm not sure of what way to go for. We don't really care about how exact (versus real time) the measurements are but rather that there is little variance in the result, since we're mostly (only?) going to use this to compare functions / code paths before and after making some changes to it.

gurka:
I just tried replacing CPUID with LFENCE, due to (from the link in my previous comment):

>So case 5 will be my go-to for now. It should have the same protection as the others (perhaps 6 is better), it is stable, and it has a low overhead and low resolution (less capacity to hide instruction latency).  Finally, the start and end measurements use the same code, which is very convenient.

I ran the same case as in my first comment and got this result:

Before:
```
 Average Time (us) | Samples | Function Name
--------------------------------------------------------------------------------
         13.749135 |    6015 | VirtioNet::msix_recv_handler()
          0.400780 |    3003 | VirtioNet::msix_xmit_handler()
--------------------------------------------------------------------------------
```

After:
```
 Average Time (us) | Samples | Function Name
--------------------------------------------------------------------------------
         13.431130 |    6013 | VirtioNet::msix_recv_handler()
          0.369620 |    3003 | VirtioNet::msix_xmit_handler()
--------------------------------------------------------------------------------
```

So, which case do we choose? lfence; rdtsc always (even if rdtscp is available) or use rdtscp if it's available, even if Barret thinks that lfence; rdtsc is more stable?

gurka:
Some more investigation. Linux uses cpu vendor and X86_FEATURE_XMM2 to select whether to use LFENCE or MFENCE before a call to RDTSC.

http://lxr.free-electrons.com/source/arch/x86/kernel/cpu/intel.c#L497
http://lxr.free-electrons.com/source/arch/x86/kernel/cpu/amd.c#L767

I propose we implement something similar. Check, using `CPUID::is_amd/intel_cpu()` and `CPUID::has_feature(XMM2)` whether to use LFENCE or MFENCE before RDTSC. Never use RDTSCP, even if it's available, per Barret's analysis ( http://akaros.cs.berkeley.edu/lxr/akaros/kern/arch/x86/rdtsc_test.c#L258 ).

fwsGonzo:
I would move all the = deleted stuff to the bottom of the class, otherwise it looks quite good :+1:

fwsGonzo:
Another thing I'm thinking about is moving stack sampler into its own header and module (stack_sampler) and keep the profiler in profile or profiler. Wanna do it?

alfred-bratterud:
Looks nice @gurka! Personally I always prefer seeing actual cycles instead of converted into time. This way you're more independent of the CPU speed.  But that's a very minor thing and just my preference :-)

rianquinn:
Back when I used to do a lot of this, I never had a nice breakdown of the different options, and depending on who you talked to (including Intel engineers), you had to use CPUID, LFENCE etc... The variance that this person is seeing is something I used to see with RDTSC as well but on older hardware. The solution that we used was case #5, but people used to swear that CPUID was needed and worked so long as you primed eax.

I think going with case #5 is reasonable as that would likely have better support in virtualized environments (to be honest your lucky it works in a VM at all). The analysis that was done was on real hardware that doesn't require a trap on that instruction so just keep in mind that because your in a VM, some of the insanity with what they guy was doing may not apply. 

The bigger question you have to ask yourself is... do you trust the results your getting 😄

alfred-bratterud:
I used to lock my qemu process to a single physical CPU using taskset when I was using `rdtsc` for a paper in order to avoid variance from CPU migrations.

gurka:
@fwsGonzo I actually prefer keeping all constructors and destructors together, especially now when we have `= delete;`. It won't make a difference to put them below `private:`, if that's what you wanted, just a different compiler error message. But if you really want them at the bottom of the class I'll put them there :)

I can split the two profilers into separate headers, but in a later PR if that's OK.

@alfred-bratterud My first version actually printed the number of cycles instead of time spent. Since we're going to use this just for comparing it doesn't really matter if we show time spent or cycles spent, so maybe I should change it back to just showing cycles. Doing that also means we could get rid of the `extern double _CPUFreq_;`, and avoid breaking stuff if we ever rename that variable. 👍

gurka:
@rianquinn Yeah, following guidelines and comments from people running tests on real hardware isn't probably the best in this case. Maybe we should conduct similar tests ourselves, within the VM, and see how much variance there really is. To find a setup that works best for us.

But for now I think that we can go with lfence / mfence; rdtsc; as it *seems* to work pretty well.

rianquinn:
Oh... agreed 100%. Unless you have a good reason to have that kind of accuracy what you have is fine. I'm actually impressed you took it this far 👍 .

fwsGonzo:
Ready to be merged?

gurka:
Yes. Would be great if someone could test this using an AMD CPU though. I've only tested it on an Intel CPU and qemu without kvm.

fwsGonzo:
I'll test this on my AMD at work whenever I get there

Issue #863, opened 2016-10-08
std::random_device crashes VM


Voultapher:

johannes:
I had the same issue. I created my own device based on C's rand() like this:

```C++
class TestDevice {
public:
  using result_type = decltype(rand);

  result_type operator()() const { return rand(); }
  double entropy() const { return 0; }
  static constexpr result_type min() { return 0; }
  static constexpr result_type max() { return RAND_MAX; }
};
```

At least in my little game library this worked nicely.

fwsGonzo:
If you want to fix it, just go to syscalls.cpp and strcomp the filename argument for `open()` towards `/dev/urandom` or `/dev/random`, then return something useful, like by using rdrand instruction.

johannes:
Well open has to return some for of file handle and then read() and others have to be implemented, along with some file handle tracking etc.

fwsGonzo:
You could make the handle for the random device(s) just some constant value :) unless that won't work

PR #893, opened 2016-10-19
Add a system PRNG, accessible through /dev/random

randombit:
Uses SHA-3's SHAKE-128 mode for the RNG. Seeding on systems with RDRAND is quite bad but better than nothing.

The `kernel/rdrand.cpp` function was disabled in the build. I added `target` annotations so it could be compiled without additional flags and added it.

Uses the approach mentioned in GH #863 of a fixed file descriptor for the rng device.

With these changes the following service works for me

```
void Service::start(const std::string&)
   {
   int fd = open("/dev/urandom", O_RDONLY);
   int random_fd = open("/dev/random", O_RDONLY);
   printf("fds %d %d\n", fd, random_fd); // the same
   assert(0 == close(random_fd)); // ignored
   uint8_t b[4096];
   assert(::read(random_fd, b, sizeof(b)) == sizeof(b));
   for(size_t i = 0; i != sizeof(b); ++i)
      printf("%02X", b[i]);
   printf("\n");
   }
```

No idea if any of this is how you'd want it, just let me know what I need to change.

### DIFF
--- a/api/kernel/rng.hpp
+++ b/api/kernel/rng.hpp
@@ -1,6 +1,7 @@
+// -*-C++-*-
 // This file is a part of the IncludeOS unikernel - www.includeos.org
 //
-// Copyright 2015 Oslo and Akershus University College of Applied Sciences
+// Copyright 2016 Oslo and Akershus University College of Applied Sciences
 // and Alfred Bratterud
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -15,29 +16,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <kernel/rdrand.hpp>
-#include <kernel/cpuid.hpp>
-#define __MM_MALLOC_H
-#include <immintrin.h>
+#pragma once
+#ifndef KERNEL_RNG_HPP
+#define KERNEL_RNG_HPP
 
-__attribute__((target("rdrnd")))
-bool rdrand16(uint16_t* result)
-{
-  int res = 0;
-  while (res == 0)
-    {
-      res = _rdrand16_step(result);
-    }
-  return (res == 1);
-}
+#include <cstdlib>
+#include <cstdint>
 
-__attribute__((target("rdrnd")))
-bool rdrand32(uint32_t* result)
-{
-  int res = 0;
-  while (res == 0)
-    {
-      res = _rdrand32_step(result);
-    }
-  return (res == 1);
-}
+// Incorporate seed data into the system RNG state
+extern void rng_absorb(const void* input, size_t bytes);
+
+// Extract output from the system RNG
+extern void rng_extract(void* output, size_t bytes);
+
+// Try to reseed the RNG state somehow
+extern void rng_reseed();
+
+// Extract 32 bit integer from system RNG
+inline uint32_t rng_extract_uint32()
+  {
+  uint32_t x;
+  rng_extract(&x, sizeof(x));
+  return x;
+  }
+
+
+#endif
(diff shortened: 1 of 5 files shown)


alfred-bratterud:
add to whitelist

fwsGonzo:
Can you redirect this to our dev branch? We are currently in the process of porting POSIX.. if it still breaks, I'll just pull this and fix it myself.

alfred-bratterud:
Thanks @randombit ! We're currently doing work on POSIX so `open`, `close` etc. are bing moved to `./posix/unistd.cpp`. Would you mind moving the modifications to those functions there and reissuing this PR to our dev branch? We'll also do something slightly more general for virtual file system support, but I think a simple hardcoded comparison of paths like you're doing now (as discussed in #863) is much better than nothing and something we can start with. When it comes to the actual crypto I don't have any expertise at all - but can you briefly educate us on how this relates to newlib's (our) `rand()` located [here](https://github.com/bminor/newlib/tree/master/newlib/libc/stdlib)?

randombit:
@fwsGonzo I will rebase against `dev` and the new interfaces, may be a day or two before I can get to it though.

The strcmp approach I used is definitely just enough of a hack for my code to work, but on the syscall implementation side it is easy to see how the internal RNG interfaces would slot into a VFS design.

On Newlibs `rand`, the first and really most essential difference is that it calls `rand48` which only has a 48-bit internal state. As a result for instance an RSA key generated by such an rng could be easily factored by simply checking all possible 48-bit seeds; this amount of computation is available to anyone with a credit card and EC2.

https://github.com/bminor/newlib/blob/master/newlib/libc/stdlib/rand48.c
https://github.com/bminor/newlib/blob/master/newlib/libc/stdlib/rand.c

In addition the algorithms used, while presumably good statistical PRNGs, are not designed for cryptographic use at all. Newlib also seems to contain an implementation of `arc4random`, which *is* a cryptographically secure RNG:

https://github.com/bminor/newlib/blob/master/newlib/libc/stdlib/arc4random.c

`arc4random` calls the `getentropy` syscall to generate the seed. I'm not sure what happens if you call it from IncludeOS, when the syscall is missing.

As to what is implemented here. The Keccak hash function (which was published in 2008 and has been intensively studied since winning the SHA-3 contest in 2012) is based on what is called a sponge function, which uses a large state and a fixed public permutation. When hashing, a piece of the input is XORed into the state, followed by running the fixed permutation. The basic idea is because the input can only affect a portion of the state and not all of it at once, an attacker cannot control the next state in any meaningful way even though the permutation is public. To produce an arbitrary length output, the sponge applies the permutation, outputs some of the bits (but never all), then repeats. Again, because part of the state (called the 'capacity') is kept back, the outputs are unpredictable.

The RNG uses the Keccak permutation in what is called duplex mode to alternately accept additional seed inputs or produce new outputs. This approach to an RNG was described by the Keccak designers in a paper some years ago (http://sponge.noekeon.org/SpongeDuplex.pdf sec 6.1).

The security of a sponge increases as the 'capacity' (amount of hidden state) increases, but the speed decreases accordingly, because fewer bytes are processed or output per permutation call. The parameters used here should provide 128 bit security.

Some additional background is https://en.wikipedia.org/wiki/Sponge_function and https://en.wikipedia.org/wiki/SHA-3

A few OSes such as FreeBSD and OS X use Yarrow or Fortuna PRNGs which are more sophisticated in how they process entropy inputs but also require a lot more code/complications. What is implemented here is basically the simplest thing that should be conservative and secure - security wise the poor seeding in the RDRAND-free case is likely the most critical issue.

alfred-bratterud:
Great writeup @randombit, thanks! I'll defer to your judgement in this matter and merge when you have time to integrate this to the dev branch. Otherwise that's something we can easily do on our side as well, just let me know.

alfred-bratterud:
@randombit we've now merged this into dev branch now (#949). We're working on a generalized solution for having strongly typed objects as entries in a virtual file system, which will probably entail creating custom file descriptors for RNG, but at least now #863 should be fixed and `std::random_device` should work from here on (integration test coming up). Thanks a lot!

randombit:
I'm glad you were able to use it, sorry about the long silence from me on this PR I have been using all available OSS dev time lately to working on Botan (incl fixing build issues for IncludeOS, should be resolved shortly via a large OCSP refactor that is currently in review).

PR #897, opened 2016-10-21
Test to prevent commits containing binary blobs

mnordsletten:
This is an experimental pull request, as I'm testing out some Jenkins functionality at the same time. I will confirm when it is ready for merging.


AndreasAakesson:
Commits binaries, who does that?!

mnordsletten:
Apparently a lot of people. The size of the github repo download is 36.7 MB. 

This is just a measure to prevent further bloat. We should ideally remove some binary files from our git history as well.

PR #899, opened 2016-10-21
test: main() + made vmrunner callbacks take trigger line as params

alfred-bratterud:


alfred-bratterud:
Blasted the commits due to binary (thanks @AndreasAakesson).

PR #911, opened 2016-10-26
Allow Debian installation

geisler:
Update installation scripts to support Debian.  Since Ubuntu is a Debian
derivative, most of the work required adding the "Debian" string to
switches that looked for Ubuntu.

The one significant change was that sudo was required to call "brctl
show".  By adding this call to the etc/create_bridge.sh script,
everything worked as expected and allowed the test.sh file to run
properly.


gurka:
Does this work on Debian stable, i.e. with clang 3.5, or are you running unstable or sid?

Also, are you sure that root really is required for `brctrl show`? The problem might be that it just can't find `brctl` since non-root users doesn't have `/usr/sbin` in their `PATH` per default. We could add the full path, but that might break the script on other systems... or just add `/usr/sbin` to `PATH` in the script?

geisler:
You're right that `brctl` can be run by the non-root user and that the path is the main problem.  Unfortunately, the path is not going to be predictable.  On my system the path is `/sbin/brctl` and not `/usr/sbin/brctl`.  Maybe the better solution is to run the entire `etc/create_bridge.sh` with `sudo` and then not have to worry about which paths belong where?  Currently, the only command that didn't require root priviledges was the `brctl show` command.

As to the version, I'm running testing and so clang-3.8 was readily available.

gurka:
Maybe we could check first if `brctl` is found in the user's `PATH` and if not then use sudo. I'm thinking of scenarios where you're using a computer that isn't yours (and don't have root/sudo access), got help to install IncludeOS the first time (work / uni IT support) and want to re-install due to e.g. newer bundle or something.

```
if ! command -v brctl> /dev/null 2>&1; then
    # use sudo brctl
    # ...
fi
```

We should state in the README that only Debian testing is supported then. Maybe even check the version of Debian the same way as UBUNTU_VERSION is checked, or something.

geisler:
The most recent commit implements the check for `brctl` as mentioned above.  I've checked it by running as my user without access to `brctl` in my path and as root with `brctl` in my path and both versions seem to work.

geisler:
I've added one more commit that definitely needs testing as I don't have jessie or Ubuntu.  The code is small, but is supposed to allow clang-3.8 to be installed on jessie since it is available through the backports repository.  Also, another test with Ubuntu should be performed to ensure that I didn't screw up that distro with this new update.

gurka:
I'll test it on jessie. I think that you'll need a `sudo` before writing to `/etc/apt` though (e.g. `sudo echo ...`, since you are not supposed to use `sudo` when executing `install.sh`. Everything that needs root access in the scripts should explicitly use `sudo`.

gurka:
I don't understand why, but ` awk 'BEGIN{ print "'$VERSION'"<"'16.04'" }'` outputs 0 for me, but `lsb_release -rs` outputs 8.5. 8.5 should be less than 16.08, right? :smile: 

I guess that we need a separate if-statement for Debian to check if `lsb_release -rs` is less than 9.0 (not 16.04), otherwise it will add backports even if you're running testing or sid. Also.. isn't `lsb_release` (about to be) deprecated?

Another way of solving this is just to state in the README that the user has to add backports manually, if running Debian jessie.

geisler:
The output of `lsb_release -rs` is `testing` and not `9.0` for stretch.  I don't know what it is for sid, but I imagine it is not numeric either.

AndreasAakesson:
ok to test

alfred-bratterud:
@geisler thanks for contributing! Could you please redirect this PR to dev branch? This way the missing test will likely pass :-)

geisler:
Not sure what the issue is since the error is in running the VM and not in the simple changes reflected by this pull request.

mnordsletten:
We've made some changes to the create_bridge.sh script, as well as to the test scripts in Jenkins. Pull in the latest from dev and take a look at the changes made to create_bridge.sh.

The changes made were to make it run on Mac OS X. The name of the bridge has been changed and a few commands were added as OS X does not have the `brctl` command.

alfred-bratterud:
retest please

fwsGonzo:
We have fixed jenkins issues now, but its no longer automatically mergable it looks like. Can you try merge dev and see if everything looks ok? If it turns green here I will merge.

geisler:
Done.

PR #912, opened 2016-10-27
binary_check test: Fix to include src path

mnordsletten:


mnordsletten:
Ok to test

mnordsletten:
Ok to test

mnordsletten:
This one is now ready for merging! Should be more robust now. 
We chose to ignore all files that have file extensions.

PR #915, opened 2016-10-28
let cmake download includeos_install and bundle all libraries within "build" directory

andreashappe:
Main goal is to make the build system self-contained, i.e., all needed libraries (included downloaded ones) should be within the build directory.

This also removes the need for external download tools.


andreashappe:
Some files still need conversion (and are still used form ~/IncludeOS_Install): crtbegin.o, crti.o crtend.o crtn.o and some include directoires (api, api/sys, mod/GSL)

andreashappe:
meh. only thing missing is libcrti and libcrtn... and those need special treatment.

andreashappe:
@alfred-bratterud , @ingve : would it be possible to merge this?

TL;DR can build includeos without ~/IncludeOS_install

This changes the cmake based build process so that:

* includeos_install will be downloaded within the build directory
* during an in-place build, i.e., make within a build directly, all are generated within the build directory, all libraries included are used from the build-directory. Linker.ld, etc. are used from within the current source or build directory
* `make install` installs libs, objects, etc. into the CMAKE_INSTALL_DIR library (slightly different directory style then ~/InludeOS_install).

alfred-bratterud:
Great @andreashappe, thanks!

PR #916, opened 2016-10-29
POSIX ucontext

MatchPointDarko:
Tell me what do you think about my implementation, i may have missed some things. anyways this code works:

     ucontext_t foo_context, foo2_context;

     void foo(int argc, size_t arg1)
     {
        getcontext(&foo_context);
        printf("argc %d\n", argc);
        printf("arg 1%d\n", arg1);
        setcontext(&foo2_context);
     } 

    void foo2()
    {
        getcontext(&foo2_context);
        printf("no args for me :(\n"); 
        setcontext(&foo_context);
    }

    void Service::start(const std::string&)
    {
        ConsoleVGA vga;
        OS::add_stdout(vga.get_print_handler());

        char buf1[1024], buf2[1024];

        getcontext(&foo_context);
        //x86 stack grows down
        foo_context.uc_stack.ss_sp = buf1 + 1024;
        foo_context.uc_stack.ss_size = 1024; 
        foo_context.uc_link = &foo2_context;
        makecontext(&foo_context, (void(*)())foo, 1, 2);

        getcontext(&foo2_context);
        //x86 stack grows down
        foo2_context.uc_stack.ss_sp = buf2 + 1024;
        foo2_context.uc_stack.ss_size = 1024; 
        foo2_context.uc_link = nullptr;
        makecontext(&foo2_context, (void(*)())foo2, 0);

        setcontext(&foo_context);
    }

Will loop forever between foo and foo2, and will print what you will think it will.


AndreasAakesson:
ok to test

AndreasAakesson:
Looks like your branch doesn't build IncludeOS successfully?

From jenkins:
```
/usr/bin/clang++-3.8 -MMD -target i686-elf -mstackrealign -msse3 -fstack-protector-strong -D_STACK_GUARD_VALUE_=1477912120 -DNO_DEBUG=1 -DOS_TERMINATE_ON_CONTRACT_VIOLATION -D_GNU_SOURCE -O2 -Wall -Wextra  -c -m32 -std=c++14 -I../api/sys -I/home/ubuntu/workspace/pull_request_stl/IncludeOS_install/libcxx/include -I/home/ubuntu/workspace/pull_request_stl/IncludeOS_install/newlib/include -Iinclude -I../api -I../api/posix -I../mod/GSL -I../mod/fiber/include -D_LIBCPP_HAS_NO_THREADS=1 -DOS_VERSION="\"v0.9.3-210-g5dba83f\"" -o kernel/os.o kernel/os.cpp
kernel/os.cpp:255:62: error: expected ')'
  default_successor_context_.uc_stack.ss_sp = (new char[1024];
                                                             ^
kernel/os.cpp:255:47: note: to match this '('
  default_successor_context_.uc_stack.ss_sp = (new char[1024];
                                              ^
1 error generated.
Makefile:204: recipe for target 'kernel/os.o' failed
make: *** [kernel/os.o] Error 1
```

MatchPointDarko:
Working on it. a minimal typo, my mistake. Compiling it again before pushing again..

alfred-bratterud:
Hi @MatchPointDarko reading over the [ucontext.h POSIX specification](http://pubs.opengroup.org/onlinepubs/009695399/functions/getcontext.html) seems to mark these functions as obsolescent, pointing to pthreads for equivalent functionality. In general don't mind having legacy functionality in the OS library, especially if they represent simpler ways to achieve things such as fibers, but I think it would be best to not have ucontext be a part of every single service, including "hello world", as it would be now (initializing the default successor context in os.cpp). Could you outline how you plan to make use of ucontext and then consider initializing a default context using a global constructor? (e.g. just move the storage of the default context to ucontext.cpp and keep it all in its own header / namespace) This way it will be available to any application that explicitly needs it.

MatchPointDarko:
@alfred-bratterud Sounds good to me, didn't know they're obsolescent. 
I think moving the default successor to ucontext.cpp is a great idea, maybe something like:
ucontext_t default_successor_context = {.uc_link = NULL, .uc_stack = {....}, .uc_mcontext = {....}};

But a code executing in global initialization needs to call 'makecontext' on this object. What if there would be a POSIX class which will initialize every POSIX component?

alfred-bratterud:
Hi @MatchPointDarko you can just use `__attribute__((constructor))` to make any C function a global constructor. See for instance [https://github.com/hioa-cs/IncludeOS/blob/master/test/kernel/integration/custom_init/custom_init2.cpp](https://github.com/hioa-cs/IncludeOS/blob/master/test/kernel/integration/custom_init/custom_init2.cpp). For code readability etc. I agree that centralized locations are nice, but I think it's important to stay as modular as possible - we want to be able to consistently say that "only what your service actually needs gets included into the binary".

MatchPointDarko:
@alfred-bratterud Ok, cool, i did not know that 👍 then i'll just create an initializing function that is a constructor, and move all the code from `OS::start` to it. Will work on it today.

alfred-bratterud:
@MatchPointDarko we're having some issues with ucontext and some questions about pthreads. Any way to get in touch with you?

PR #918, opened 2016-10-31
WIP: Work on external builds (3rd party unikernels)

andreashappe:
I'm trying to make icrd compile as external project. So far it works except libvirtnet.a support (as now this is a libarry, not an object file).

How to: from within checkout-out includeos project:

~~~ bash
$ mkdir build
$ cd build
$ CMAKE_INSTALL_PREFIX=~/tmp CXX=/usr/bin/clang++ CC=/usr/bin/clang cmake ./..
$ make
$ make install
~~~

Now everything will be installed into ~/tmp , including a custom Makeseed

Then checkout ircd and change its Makefile  (mostly adapt path to Makeseed file):

~~~
iff --git a/Makefile b/Makefile
index fe9eb34..27f77c7 100644
--- a/Makefile
+++ b/Makefile
@@ -14,18 +14,18 @@ FILES = service.cpp ircd.cpp client.cpp client_new.cpp client_msg.cpp client_cmd
 DISK=
 
 # Drivers to be used with service
-DRIVERS=virtionet
+DRIVERS=
 
 # Your own include-path
 LOCAL_INCLUDES=
 
 # IncludeOS location
 ifndef INCLUDEOS_INSTALL
-INCLUDEOS_INSTALL=$(HOME)/IncludeOS_install
+INCLUDEOS_INSTALL=$(HOME)/tmp/
 endif
 
 # Include the installed seed makefile
-include $(INCLUDEOS_INSTALL)/Makeseed
+include $(INCLUDEOS_INSTALL)/share/includeos/Makeseed
 
 # Disable crash context buffer
 #CPPOPTS += -DDISABLE_CRASH_CONTEXT
~~~

`make` should do it now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,8 +26,12 @@ if(examples)
 	add_subdirectory(examples)
 endif(examples)
 
-install(DIRECTORY api/ DESTINATION include/api
-	FILES_MATCHING PATTERN "*.h")
+install(DIRECTORY api/ DESTINATION include/api)
 
 enable_testing()
 add_subdirectory(test)
+
+configure_file(etc/Makefile.seed.in Makeseed)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/Makeseed DESTINATION share/includeos)
+
+install(DIRECTORY mod/GSL/gsl DESTINATION include/gsl)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,3 +35,8 @@ configure_file(etc/Makefile.seed.in Makeseed)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/Makeseed DESTINATION share/includeos)
 
 install(DIRECTORY mod/GSL/gsl DESTINATION include/gsl)
+
+set(CPACK_GENERATOR "TGZ;DEB")
+set(CPACK_PACKAGE_VERSION ${OS_VERSION})
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Ingve")
+include(CPack)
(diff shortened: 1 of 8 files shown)


andreashappe:
@alfred-bratterud any good idea how to include the driver libraries? I was thinking about maybe just adding a $(ADDITIONAL_LIBRARIES) variable but I'm not sure if that's enough.

alfred-bratterud:
@andreashappe thanks! I'm a little worried about duplicating the seed Makefile as this will give us yet another file to maintain. Wouldn't it be better to have all projects use CMake than to mix the two? This would afaict also solve the driver inclusion issue no?  e.g. have ./seed contain CMakeLists.txt and direct people to edit that one when they copy seed to produce their own service. Thoughts?

andreashappe:
@alfred-bratterud : I can understand that, but IMHO this shouldn't be too much of a problem With cmake I see two different ways of helping external projects:

* configure_file (as used with Makefile.seed): this is mostly a templating engine and allows to dynamically fill in data (to reduce redundancy). I've used this to dynamically generate a Makeseed file using the (user-configured) library directories. In the long term also C-flags, etc. could be configured this way.
* export_file -- should export a CMakeLists.txt that an external project itself can import. This should allow to use use all cmake objects (such as libos) within an external project. Haven't tried this one  yet.

This is just an initial step. My cmake is still rusty, so I'm fully convinced that other people will further cleanup and streamline the installation process.

andreashappe:
@alfred-bratterud : could you cherry pick 9e2e2fa and 1251fbb (or merge the whole branch into dev)? I added those two commits after @mnordsletten merged my pull-request into his

fwsGonzo:
test please

PR #920, opened 2016-11-03
Cmake and unit tests now seperate tests

mnordsletten:


alfred-bratterud:
@mnordsletten @andreashappe I guess this will merge the Cmake stuff with the duplicate makefile that we want to avoid, but we can fix that later. We do want a cmake system for external projects. Ideally I think we should have a CmakeLists.txt in ./seed, and have Jenkins copy seed to an external location and make sure it builds and runs. @andreashappe how would you do this?

andreashappe:
TBH I think, that a single makefile is not too much redundancy. But we can easily create another CMakeFile in a similar fashion (configure_file) or try the export() approach (https://cmake.org/cmake/help/v3.0/command/export.html). Will look into it soon -- not sure what is the better way yet.

andreashappe:
@alfred-bratterud : another option would be a includeos-config binary that would be automatically be installed, similar to, e.g., the libbpng15-config tool:

~~~
% libpng16-config --help
Usage: /usr/bin/libpng16-config [OPTION] ...

Known values for OPTION are:

  --prefix        print libpng prefix
  --libdir        print path to directory containing library
  --libs          print library linking information
  --ccopts        print compiler options
  --cppflags      print pre-processor flags
  --cflags        print preprocessor flags, I_opts, and compiler options
  --I_opts        print "-I" include options
  --L_opts        print linker "-L" flags for dynamic linking
  --R_opts        print dynamic linker "-R" or "-rpath" flags
  --ldopts        print linker options
  --ldflags       print linker flags (ldopts, L_opts, R_opts, and libs)
  --static        revise subsequent outputs for static linking
  --help          print this help and exit
  --version       print version information

% libpng16-config --ldflags
-L/usr/lib64 -lpng16
~~~

if the includeos-config tool would offer switches for includes, libs, ld-flags, etc. as well as switches as --enable-virtionet , that might work too.

PR #924, opened 2016-11-04
net: Combined Link layer into physical driver

AndreasAakesson:
In other words; Driver now decides what link layer to use. To build a stack you no longer need to route link to physical, this is done by the driver when Nic is created. Nic now returns a delegate to the link. 

When knowledge of what link layer is used is necessary (IP4 <-> ARP <-> Ethernet), enum `Proto` in Nic can tell.


AndreasAakesson:
retest please

AndreasAakesson:
* Fixed calculation of ICMP length - no longer result in incorrect checksum (on Mac, didnt show on Linux)
* Size is now correctly set on incoming packets in Virtionet (subtracted vnet header from packet size)

AndreasAakesson:
retest please

PR #936, opened 2016-11-16
Add docker based image builder / runner

xwvvvvwx:
This allows the following workflow to run a new IncludeOS service:

```
$ cp -r seed/ <service_path>
$ cd <service_path> && ./docker_run.sh
```

The `./docker_run.sh` script passes arguments through to docker (if no arguments are passed, it will call `run.sh`):
    - to build an image (without running): `./docker_run.sh make`
    - to start a new bash session: `./docker_run.sh bash`

Two unresolved issues:

1. I couldn't figure out how to bind ports on a running service to the docker host. This means that if you want to `curl` / `ping` the service, you need to start a bash shell in the container and build / test inside the container. If someone knows how to get this working, I would be more than happy to improve this.

2. `./docker_run.sh` assumes that the image is tagged with `includeos`. This tag is not present in the docker hub right now, so the command will fail if such an image is not uploaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,8 @@
 *.ropeproject
 nbproject
 
+!Dockerfile
+
 src/apic_boot.bin
 src/bootloader
 src/TAGS
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM ubuntu:xenial
+
+RUN apt-get update && apt-get -y install \
+    git \
+    net-tools \
+    sudo \
+    wget \
+&& rm -rf /var/lib/apt/lists/*
+
+RUN useradd --create-home -s /bin/bash ubuntu
+RUN adduser ubuntu sudo
+RUN echo -n 'ubuntu:ubuntu' | chpasswd
+
+# Enable passwordless sudo for users under the "sudo" group
+RUN sed -i.bkp -e \
+      's/%sudo\s\+ALL=(ALL\(:ALL\)\?)\s\+ALL/%sudo ALL=NOPASSWD:ALL/g' \
+      /etc/sudoers
+
+USER ubuntu
+
+ADD . /home/ubuntu/IncludeOS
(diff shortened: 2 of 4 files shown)


AndreasAakesson:
add to whitelist

alfred-bratterud:
Hi @xwvvvvwx thanks for contributing! I think @ingve is the right person to review this request, I know we are using docker to build IncludeOS in Unik.

ingve:
Very nice! Favourite parts: the separate step for setting up the bridge, and adding a user. :) 

As @alfred-bratterud mentions, we use Docker to build IncludeOS in UniK, but in our existing Dockerfile (distributed with UniK) the design goals were to make as few changes as possible to the install script, keep everything IncludeOS-related *inside* the container (the IncludeOS development environment should *not* be installed locally on the host), and to use the specific paths that UniK expects:

https://github.com/emc-advanced-dev/unik/tree/master/containers/compilers/includeos/cpp

Your Dockerfile is a bit nicer/more general, and it would be good if we distributed a stand-alone Dockerfile in our own repo (and on Docker Hub) as well, for those who want a Docker-based workflow, but we should try to make sure that the "building the service" part can be extracted into its own Dockerfile so we can reuse it in the UniK compiler. Then the "running the service" Dockerfile can build on top of that.

xwvvvvwx:
Glad you like it :)

Right now I use the same docker image for building and running the services, the only difference is that to build you need to pass `make` as the command, and for running you pass `run.sh` as the command to docker.

I just had a quick skim through the Dockerfile in the unik repo, and I think that this [Dockerfile.hw](https://github.com/emc-advanced-dev/unik/blob/master/containers/compilers/includeos/cpp/Dockerfile.hw) could just pull from the official includeos docker image, instead of the [Dockerfile.common](https://github.com/emc-advanced-dev/unik/blob/master/containers/compilers/includeos/cpp/Dockerfile.common) in the Unik repo?

PR #941, opened 2016-11-17
api: Add type alias for observable pointer

RicoAntonioFelix:


alfred-bratterud:
@RicoAntonioFelix thanks for contributing! Could you please explain the rationale? I'm a bit concerned that this will just convolute the fact that we're using raw pointers. I did use an Obs_ptr alias in VFS, but that was while experimenting with various smart pointers. I ended up returning a raw pointer some places due to efficiency, but the idea was that they should be for internal use only. Are there any compile time advantages to wrapping the pointer in a template?

RicoAntonioFelix:
Hey @alfred-bratterud...The reasoning of this PR was to introduce a vocabulary type within the project to convey the meaning of an observable pointer, one that doesn't concern itself with lifetime management but just a handle to some object for the purpose of observing its state...

That way over time working with the codebase using `handle<T>` you don't have to think about what it mean as opposed to `T*` which can have multiple meanings...

Since it's just a type alias it doesn't introduce any runtime overhead...

alfred-bratterud:
I see @RicoAntonioFelix, @ingve points out that this resembles [std::experimental::observer_ptr](http://en.cppreference.com/w/cpp/experimental/observer_ptr) - maybe we should go with that instead?

mnordsletten:
I'm investigating why the Network test timed out. Jenkins test please

RicoAntonioFelix:
@alfred-bratterud I would have gone with a standard component after he pointed it out to me as well, but unfortunately it doesn't exist within our library...

ingve:
@RicoAntonioFelix is absolutely right, observer_ptr and the other Library Fundamentals 2 parts are [not part of Clang yet](http://clang.llvm.org/cxx_status.html) but there are other standards-compliant implementations we could investigate, for example Martin Moene (the author of Lest) has [observer_ptr<> for C++98 up](https://github.com/martinmoene/observer-ptr).

alfred-bratterud:
How about just renaming `handle` to `observer_ptr` and then changing our `using` alias when clang gets updated with the new standard? This way we'd get the semantics right and keep the API compatible with C++ 17? Looks like dereferencing an observer holding nullptr is undefined behaviour so no extra type safety anyway.

RicoAntonioFelix:
To me the only benefit is the point of adding vocabulary to code which enhances readability and by extension maintenance...If you have __something__ that is used to observe $$something$$ else one must make sure that __something__ is in a valid state to perform observation...

fwsGonzo:
We are in the process of merging in clang build system, did it interfere with your changes?

RicoAntonioFelix:
My local repo got messy so I decided to start from a clean slate...

PR #956, opened 2016-11-23
Build services with CMake on Mac

AndreasAakesson:
By using the same toolchain, services can now be built on Mac.

`cmake -DCMAKE_TOOLCHAIN_FILE=/usr/local/includeos/mac-toolchain.cmake`

It currently assumes binutils tools are installed with the correct prefix. Will update the script to install this correctly.


AndreasAakesson:
Made `install_osx.sh` a "dependency installation" script. Work as before, but now installs binutils in a way so it can be used with `mac-toolchain.cmake` (in `/usr/local/bin`).
Script has to be run from root now tho, since mac script is using `install_binutils.sh` script.

AndreasAakesson:
da1553248 hopefully fixes the error on linux

PR #961, opened 2016-11-25
demo_service example rework

Voultapher:
My goals:
Show that it is more than a static website
Show it can have state across multiple sessions
Make it fun
Make it pretty


ingve:
Make it fun and make it pretty, that's what it's all about!

Hey Jenkins, this is ok to test

ingve:
ok to test

mnordsletten:
Most of these tests are failing due to us having just moved to cmake. We are still in the process of integrating all these tests with cmake. Just give us a few days and this should all be up and running again.

AndreasAakesson:
@Voultapher Merge with latest dev when you have the time and let jenkins run the tests again :)

alfred-bratterud:
jenkins test please

alfred-bratterud:
Hi @Voultapher the project looks really nice :-) My main concern is that we probably don't want to encourage people to do manual http handling but rather use a library like [http](https://github.com/includeos/http) or a framework like [mana](https://github.com/includeos/mana). That being said, I'd like to get this merged. As a first step, how about: 1) moving the html to a separate `.html`-file residing on a memdisk and 2) writing a few sentences in the README pointing out that this service wants to show what you can do with only the core repository and point to the other projects mentioned above, including [Acorn](https://github.com/includeos/acorn), for more full featured solutions? When we end up either using a package manager or merging http into the core we can rework your new example to use that. Any thoughts? Thanks for putting in the work here, it's much appreciated :-)

Voultapher:
I changed my fork, so I'll reopen a new pr.

PR #973, opened 2016-12-03
Added non storage assuming function new_memdisk()

Voultapher:


AndreasAakesson:
add to whitelist

alfred-bratterud:
Hi @Voultapher, in principle I don't mind having static storage for the disk object itself, but it might be a bit strange to have to instances of a file system for the same block device.  We'll also be integrating the memdisk into the virtual file system (vfs) that's in progress so that you'll eventually have a single plugin that mounts all the disks found on the system to one central location. If there are no objections from @fwsGonzo I'll happily merge this for now - I'll see if I can get the tests to pass. Jenkins test please.

Voultapher:
wrong branch

PR #974, opened 2016-12-04
added check for lsb_release in install script

szborows:
I had no `lsb_release` installed on my system and `install.sh` complained the system was not supported at all.

### DIFF
--- a/etc/install_build_requirements.sh
+++ b/etc/install_build_requirements.sh
@@ -13,10 +13,10 @@ case $SYSTEM in
         ;;
     "Linux")
         case $RELEASE in
-            "Debian"|"Ubuntu"|"LinuxMint")
+            "debian"|"ubuntu"|"linuxmint")
                 VERSION=`lsb_release -rs`
                 if [ $(awk 'BEGIN{ print "'$VERSION'"<"'16.04'" }') -eq 1 ]; then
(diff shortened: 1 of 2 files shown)


rianquinn:
@alfred-bratterud FYI... we had the same problem with lsb_release. We started using /etc/os-release instead which seems to be on most modern Linux distros.

szborows:
retrieving this manually is a little bit trickier in `sh` than I initially though. apparently `sh` doesn't support arrays like `bash` does.
what do you think about following? can we assume we have `grep` and `cut` on the board?

```sh
grep "ID=" /etc/os-release | cut -c 4-
```

alfred-bratterud:
Hi @szborows, thanks for contributing! Looks like both grep and cut are listed in [posix utilities](http://pubs.opengroup.org/onlinepubs/9699919799/idx/utilities.html), so those should be fine. I'll see if I can get the tests to pass. Jenkins test please.

szborows:
actually there's a way to do this without cut:
```sh
LINE=`grep "ID=" /etc/os-release`
RELEASE=`echo "${XYZ##*=}"`
```

badum-tsss

szborows:
are these tests inherently unstable?

alfred-bratterud:
jenkins test please

PR #975, opened 2016-12-05
Booting with GRUB

alfred-bratterud:
Fixed stuff required to boot with GRUB (refreshing segment regs from GDT) and added an automated test to build and boot a grub-based IncludeOS image.


alfred-bratterud:
jenkins test please

PR #978, opened 2016-12-06
Reading tar file

AnnikaH:


AnnikaH:
jenkins test please

AnnikaH:
cmake and misc fail can be ignored - submodule uzlib should be added (ignore binary file warning)

etc/service.cmake
```diff
@@ -221,6 +221,39 @@ foreach(DISK ${MEMDISK})
   target_link_libraries(service --whole-archive disk${MDCOUNTER} --no-whole-archive)
 endforeach()
 
+if(TARFILE)
+  get_filename_component(TAR_RELPATH "${TARFILE}"
+                         REALPATH BASE_DIR "${CMAKE_SOURCE_DIR}")
+
+  if(CREATE_TAR)
```
AndreasAakesson: Maybe all this could be replaced with a CMake function instead. The function could either set `TARFILE`, or only work if `TARFILE` is already set. Here's a sketch for the former:
```
function(create_tar DEST SRC)
  set(TARFILE ${DEST})
  get_filename_component(TAR_RELPATH "${TARFILE}"
                         REALPATH BASE_DIR "${CMAKE_SOURCE_DIR}")
  get_filename_component(TAR_EXT "${TAR_RELPATH}" EXT)

  if(TAR_EXT strequal ".tar.gz")
    COMMAND tar czf ${TAR_RELPATH} ${SRC}
  else(TAR_EXT)
    COMMAND tar cf ${TAR_RELPATH} ${SRC}
  endif(TAR_EXT)
endfunction(create_tar)
```

(I haven't tried it out tho)

alfred-bratterud:
jenkins test please

PR #979, opened 2016-12-06
net: Added HTTP functionality

RicoAntonioFelix:


AndreasAakesson:
Looks like binary_check rejects adding new submodules. @mnordsletten any ideas?

mnordsletten:
I'll look into that when I get back. I'm adding it to my list.

PR #980, opened 2016-12-06
fix udp listen port to 9967

ilackarms:
due to conflicts with another application, we've changed the udp broadcast port in UniK instance listener


ingve:
add to whitelist

AndreasAakesson:
jenkins test please

PR #984, opened 2016-12-08
Add optional profilers to OS start, fix issue where ScopedProfiler constructor could be moved around

fwsGonzo:
Also move the initial call to OS::event_loop from OS::start to kernel_start to allow running scoped profilers in OS::start

### DIFF
--- a/api/kernel/os.hpp
+++ b/api/kernel/os.hpp
@@ -144,6 +144,9 @@ public:
   static void block();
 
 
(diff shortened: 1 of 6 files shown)


fwsGonzo:
Still retaining profilers in OS::Start, however its no longer necessary to have one active profiler for a fast soft-reset as it seems to have been a nast side effect of the PCI hardware deactivation function not working correctly..

PR #993, opened 2016-12-09
HTTP Client, TCP improvements and update GSL

AndreasAakesson:
* Simple randomization of ports (got really tired of all resets because of the same outgoing port being used when service got restarted a lot)
* Some minor changes to TCP, makes it possible to correctly write (queue) and hook up on read before connection has been established.
* HTTP Client (WIP)
* Update GSL to latest version

### DIFF
--- a/src/net/http/connection.cpp
+++ b/src/net/http/connection.cpp
@@ -44,7 +44,6 @@ namespace http {
   void Connection::send(Request_ptr req, Response_handler on_res, const size_t bufsize)
   {
     req_ = std::move(req);
-    res_ = std::make_unique<Response>();
     on_response_ = std::move(on_res);
 
     send_request(bufsize);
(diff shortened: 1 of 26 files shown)


src/net/http/connection.cpp
```diff
@@ -0,0 +1,87 @@
...
+  void Connection::recv_response(buffer_t buf, size_t len)
+  {
+    Expects(res_ != nullptr);
+    *res_ << std::string{(char*)buf.get(), len};
+    res_->parse();
+
+    const auto& header = res_->header();
+    // TODO: Temporary, not good enough
+    // if(res_->is_complete())
+    if(!header.is_empty() && std::stoi(header.value(header::Content_Length).to_string()) == res_->body().size())
```
AndreasAakesson: Not sure how to solve the buffering yet.

AndreasAakesson:
There is also a fix for binary check e3216de

PR #997, opened 2016-12-12
diskimagebuild: skeleton

ingve:
(Not strictly a real PR, just trying to nudge @fwsGonzo to create a disk image builder that works for those of us building on MacOS now that everything else works there...) :)


mnordsletten:
Jenkins test please

mnordsletten:
Jenkins test please

PR #1003, opened 2016-12-13
Only install packages if required on Arch Linux

SuperV1234:
More info here: http://superuser.com/questions/568967/make-pacman-does-not-reinstall-when-already-installed


alfred-bratterud:
Hi @SuperV1234 thanks for contributing! Can you please direct this PR to our dev branch?

alfred-bratterud:
Oh, just noticed that this seems to overlap with #1006 which already points to dev

PR #1004, opened 2016-12-13
Fix 'using declaration cannot refer to a scoped enumerator' error

SuperV1234:
Instead of `using`, `constexpr auto` can be used to alias the `enum class` values.

### DIFF
--- a/src/kernel/cpuid.cpp
+++ b/src/kernel/cpuid.cpp
@@ -39,10 +39,10 @@ namespace
 
   constexpr auto get_feature_info(Feature f)
   {
-    using Register::EAX;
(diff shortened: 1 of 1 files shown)


fwsGonzo:
add to whitelist

SuperV1234:
Ah, it was already fixed in dev by explicitly qualifying all values. I think this can be closed unless you want to fix the current master.

fwsGonzo:
We are close to merging master again. In less than a week or so, so probably not worth the effort in merging across the branches again. :) Do keep any PRs that contain fixes not present on dev though.

PR #1023, opened 2016-12-17
uri: Optimize module

RicoAntonioFelix:
# Change 
* The underlying `uri` parser to use [http-parser](https://github.com/nodejs/http-parser)
* Queries to use `string_view` to create sub-views into the given `uri`


AndreasAakesson:
jenkins test please

PR #1031, opened 2016-12-22
Fix compile errors when building with clang v3.9+

moncruist:
There was a bug in clang which accepts the aliasing of scoped enumerator through 'using' keywords. In clang v3.9 this bug was fixed and that leads to compile errors when you try to build master with new clang. This PR fixes that errors.

### DIFF
--- a/api/net/packet.hpp
+++ b/api/net/packet.hpp
@@ -102,12 +102,8 @@ namespace net
     { return chain_.get(); }
 
     /* Get the tail, and detach it from the head (for FIFO) */
-    auto detach_tail() noexcept
-    {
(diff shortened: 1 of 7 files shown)


ingve:
ok to test

ingve:
Hi! Thanks for this! Could you redirect this PR to the `dev` branch? (`dev` should get merged into `master` Real Soon Now™)

moncruist:
Oh. It seems there are similar fixes in dev branch (9452edb0841b5734939d19ea8e872dd0dd4effd8). I think this PR should be closed as outdated.

PR #1043, opened 2017-01-03
Changes needed to compile SQlite service

fwsGonzo:
.. and
* Kernel sanity checks
* kprint now works in C modules
* Make the string parameter for Service::start optional


fwsGonzo:
jenkins test please

Issue #967, opened 2016-11-30
Failed tests does not cleanup


AndreasAakesson:
When I run `test/test.py` there is some tests failing due to missing lest (#966) and running on Mac (linux specific commands #932). These all leave an unfinished build `<test>/build`. I guess this is not an issue when working on a single test, but it's a lot of work to clean up every failed test.

Should every test clean up on exit?
Or maybe have the root test.py clean up every test recursively (with supplied arg or when done)?

alfred-bratterud:
There's an option `./test.py -c` that will clean up each test before running a second time.

AndreasAakesson:
Thanks! Tho I guess this will end up with the same result since the test is ran once again. I guess I can add a second option for just cleaning, or some flag that breaks after the cleaning is done. Any suggestions?

PR #1049, opened 2017-01-09
testrunner: Dynamic number of parallell tests based on cpus

mnordsletten:
- Now runs as many tests in parallell as number of cpus
- Added argument to testrunner.py which only cleans the tests


mnordsletten:
Does previous commit solve #967 @AndreasAakesson ?

AndreasAakesson:
Yupp, nice :)

mnordsletten:
ok to test

PR #1052, opened 2017-01-09
Various fixes to compile with 64bit G++, move handlers in Timers

fwsGonzo:
DHCP now also throws exception when not initialized before calling on_config. Not sure if it's the right thing to do.


mnordsletten:
Jenkins test please

mnordsletten:
Jenkins test please

PR #1057, opened 2017-01-10
vmrunner/boot: skip disk drive if only kernel is provided + cosmetic …

alfred-bratterud:
…changes


alfred-bratterud:
This is supposed to make it possible to do `$ boot seed` in the current `./seed` folder and it should just work.

PR #1059, opened 2017-01-12
Support for test-coverage shields

mnordsletten:
In order to obtain proper shields with test coverage we needed a junit implementation within our testrunner.py script. To generate `output.xml` file used by jenkins (and shields.io) use the `-j` option. 

[Example](https://jenkins.includeos.org/view/All/job/shield_mnordsletten/)

shield-here

This can enable:
- A clearer split between build status, and test status in the readme.
- All pull requests can report how many tests actually failed. 

Required some small rewrites of the testrunner.py. It simplifies the way tests are run, by adding both stress test and the misc tests to be discovered at an earlier stage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 *.d
 *.pyc
 *.ropeproject
+*.xml
 nbproject
 
 !Dockerfile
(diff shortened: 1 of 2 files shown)


mnordsletten:
Had to update all build nodes with new python dependency: `junit-xml`

jenkins test please

PR #1062, opened 2017-01-13
Tar and examples update

AnnikaH:
const, noexcept, cleanup examples incl. run.sh update, removed Makefiles and updated READMEs (except STREAM and 256_color_vga)


AndreasAakesson:
jenkins test please

mnordsletten:
I did see that Jenkins ran the test after you commented andreas. Though instead of triggering the latest commit "cb35b2c" it ran the previous commit that was tested again: c06032d

I will try it again to see if this error repeats itself.

jenkins test please

PR #1074, opened 2017-01-20
Use Inet<net::IP4> instead of Inet4

themodder666:


AndreasAakesson:
Thanks @themodder666. Can you please change the include path as well?: `#include <net/inet4>` => `#include <net/inet>`

themodder666:
done

themodder666:
i see its been done elsewhere. thanks

Issue #1073, opened 2017-01-20
Change Mana Server to accept net::Inet<net::IP4>


themodder666:
Currently,the mana server accepts net::Inet4 as the IP_stack. This should be changed to net::Inet<net::IP4> to have better compatibility with ifconfig and be more general.

AndreasAakesson:
That's a good idea, thanks!

PR #1076, opened 2017-01-20
HTTP Client improvements, DNS Client rework, RFC 6335

AndreasAakesson:
For some reason the the callback delegate the HTTP Client sometimes ended up as nullptr. Maybe due to new delegate implementation. Seems to be fixed by using `spec::dynamic`. If anyone can explain why, please do :D

TCP and UDP are now using dynamic (ephemeral) ports in the range described by IANA in [RFC 6335](https://tools.ietf.org/html/rfc6335). In other words, the outgoing ports are now in range of `49152 - 65534`.

Improved the DNS Client by adding timeout on resolve requests (5s default). It now also reuses one UDP socket. Also added transaction ID matching on incoming responses.

Also a fix for #1073 sneaked in before I noticed #1074 - please merge that one first (when sorted).

### DIFF
--- a/api/net/dns/dns.hpp
+++ b/api/net/dns/dns.hpp
@@ -151,18 +151,21 @@ namespace net
     public:
       int  create(char* buffer, const std::string& hostname);
       bool parseResponse(const char* buffer);
-      void print(char* buffer);
+      void print(const char* buffer) const;
 
       const std::string& getHostname() const
       {
         return this->hostname;
       }
+
       IP4::addr getFirstIP4() const
       {
-        if (answers.size())
-          return answers[0].getIP4();
+        auto it = answers.begin();
 
-        return IP4::ADDR_ANY;
+        while(it != answers.end() and !it->is_answer_type(DNS_TYPE_A))
+          ++it;
+
+        return (it != answers.end()) ? it->getIP4() : IP4::ADDR_ANY;
       }
 
     private:
@@ -175,7 +178,10 @@ namespace net
         rr_data resource;
 
         IP4::addr getIP4() const;
-        void      print();
+        void      print()  const;
+
+        bool is_answer_type(int type) const
+        { return ntohs(resource.type) == type; }
 
       private:
         // decompress names in 3www6google3com format
--- a/api/net/dns/dns.hpp
+++ b/api/net/dns/dns.hpp
@@ -162,10 +162,14 @@ namespace net
       {
         auto it = answers.begin();
 
-        while(it != answers.end() and !it->is_answer_type(DNS_TYPE_A))
+        while(it != answers.end())
+        {
+          if(it->is_type(DNS_TYPE_A))
+            return it->getIP4();
           ++it;
+        }
 
-        return (it != answers.end()) ? it->getIP4() : IP4::ADDR_ANY;
+        return IP4::ADDR_ANY;
       }
 
     private:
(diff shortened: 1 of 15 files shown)


src/net/tcp/tcp.cpp
```diff
@@ -131,7 +131,7 @@ seq_t TCP::generate_iss() {
 */
 port_t TCP::next_free_port() {
 
-  current_ephemeral_ = (current_ephemeral_ == 0) ? current_ephemeral_ + 1025 : current_ephemeral_ + 1;
+  current_ephemeral_ = (current_ephemeral_ == port_ranges::DYNAMIC_END) ? port_ranges::DYNAMIC_START : current_ephemeral_ + 1;
 
   // Avoid giving a port that is bound to a service.
   while(listeners_.find(current_ephemeral_) != listeners_.end())
```
RostakaGmfun: I think the below line (`current_ephemeral_++;`) contains a bug. The value of `current_ephemeral_` may overflow the `port_ranges::DYNAMIC_END` if the last port from the dynamic port range is used by the listener. Am I right?

src/net/dns/client.cpp
```diff
@@ -17,35 +17,49 @@
...
 {
+  Timer::duration_t DNSClient::DEFAULT_RESOLVE_TIMEOUT{std::chrono::seconds(5)};
+
   void DNSClient::resolve(IP4::addr dns_server, const std::string& hostname, Stack::resolve_func<IP4> func)
   {
-    auto& sock = stack.udp().bind();
-
     // create DNS request
     DNS::Request request;
     auto*  data = new char[256];
```
RostakaGmfun: Could we allocate the `data` on stack if the size is known at compile time?

api/net/dns/dns.hpp
```diff
@@ -149,22 +149,33 @@ namespace net
...
       const std::string& getHostname() const
       {
         return this->hostname;
       }
+
       IP4::addr getFirstIP4() const
       {
-        if (answers.size())
-          return answers[0].getIP4();
+        auto it = answers.begin();
```
RostakaGmfun: Have you considered using range-based for loop?

src/net/dns/client.cpp
fwsGonzo: Seems like it to me

AndreasAakesson:
Thanks for the review @RostakaGmfun, much appreciated!

src/net/dns/client.cpp
AndreasAakesson: Yeah didn't think about that one.

api/net/dns/dns.hpp
AndreasAakesson: Guess I somehow overlooked it, good suggestion.

src/net/tcp/tcp.cpp
AndreasAakesson: I think you're right here yes. But it shouldn't be a listener on one of the dynamic ports (we do not prevent this tho, and I'm not sure if we want to do this either?). I know the TCP port-handling is not bulletproof, and that it maybe should be reworked some. 
What makes it a bit troublesome is that there are two collections where a port could be taken; `listeners_` (listening bound ports) and `connections_` (outgoing connections).
Another bug is when all dynamic ports are taken/active, we start overwriting existing ones. The port-handling in UDP is much more simpler/safer.

PR #1092, opened 2017-01-25
boot: behaves more like you'd expect (WIP)

alfred-bratterud:
* No cmake required for booting binaries
* Added json option to pass arbitrary config
* Use verbose option in vmrunner
* Better defaults

Importantly now all of these should work:
* `$ boot .`, or `$ boot any/test/or/example` requires CMakeLists.txt as before
* `$ boot path/to/any_image.img`, no cmake required, json-config from current folder if exists
* `$ boot path/to/any_binary` 
* `$ boot --config path/to/my.json path/to/any_binary`, no cmake required, selected json-config


AndreasAakesson:
jenkins test please

alfred-bratterud:
The failing tests used to never get run due to being incomplete. After having relaxed the requirements for what you can boot they now get run. Solution is probably to have separate and more strict validation of tests than of just vm's.

gurka:
Does this make the example in the root README.md work? e.g. this one

```
    $ cp -r seed ~/my_service
    $ cd ~/my_service
    $ emacs service.cpp
    ... add your code
    $ mkdir build && cd build
    $ cmake ..
    $ make
    $ boot my_service
```

Currently, on master, the above doesn't work since `boot` can't find a `CMakeLists.txt` in the current directory (build). Or perhaps this is how it should work and the problem is a missing step in the example (`$ cd ..` in between `$ make` and `$ boot my_service`)?

fwsGonzo:
These are improvements that solve issues such as that. The intention is that boot should just work and people shouldn't have to ask about how it works.. ideally, I guess. So if you get the current dev, and boot works unexpectedly that would be good feedback :+1:

PR #1107, opened 2017-01-31
Removed unused variable

jsfdez:


jsfdez:
jenkins test please

ingve:
jenkins test please

ingve:
Thank you for trying IncludeOS and for contributing! 

The current `fpathconf` code is pretty ugly. The call to `FD_map::_get` might seem superfluous and the variable might seem unused, but the function is used for its side effect (it throws if the file descriptor is not found in the map, and the catch block sets `errno` to `EBADF` (Bad file descriptor)).

According to the [POSIX Specification](http://pubs.opengroup.org/onlinepubs/9699919799/), "The fpathconf() function **shall** fail if [...t]he fildes argument is not a valid file descriptor." (And according to the description of the function, "the fildes argument is an **open file descriptor**"). (Emphasis mine.)

If I understand the code in your PR correctly, by ignoring the fd argument, it always returns a value, and in the case where the fd argument does not refer to an open/valid file descriptor, it never fails as required by the specification?

alfred-bratterud:
Yea good intentions but I agree that the check has necessary side-effects. Please feel free to suggest further improvements - POSIX is a large project and there's a lot left to do.

PR #1129, opened 2017-02-06
Improve Arch Linux compatibility

SuperV1234:
* Use `env python2` instead of `env python` in the `boot` script
* Install required `python2 python2-jsonschema python2-psutil` packages
* Explicitly add `clang-3.9` support


ingve:
jenkins test please

PR #1133, opened 2017-02-09
test: update mod/gsl test to work

alfred-bratterud:


mnordsletten:
I enabled the test again. Let's see if it passes.
jenkins test please

PR #1137, opened 2017-02-10
util: Add support for encoding and decoding C-String's in base64 module

RicoAntonioFelix:


fwsGonzo:
Does the std::vector constructor make a copy of the data here, before encoding it?

RicoAntonioFelix:
As you mentioned that, I refactor it...

PR #1145, opened 2017-02-13
kernel: Make elf parser use no-heap only, update stack sampler

fwsGonzo:
... and add print function


fwsGonzo:
jenkins test please

PR #1147, opened 2017-02-14
icmpv4: Remove <os> include & fix broken include path

dfct:
#include <os> is extraneous here, we don’t need to take on the
dependency. The icmpv4.hpp is short a ../ in its relative pathing, but
regardless should use <net/ip4/icmpv4.hpp> directly instead. This
mirrors the styling in other ip cpp files.


ingve:
jenkins test please

PR #1148, opened 2017-02-14
src/net - Remove old inet.cpp file (replaced by inet4.cpp)

dfct:
Inet.cpp was replaced by inet4.cpp and is no longer used in builds nor
otherwise compilable.


fwsGonzo:
add to whitelist

PR #1156, opened 2017-02-14
net - Update old debug print variables

dfct:
Updated old debug print variables such that the code compiles in debug again. (This time without introducing an unused variable warning in release mode :)


fwsGonzo:
We just updated with new changes to make the TAP devices faster (reaching gigabits), so conflicts apparently!

dfct:
Haha, nice! A couple of the prints I was fixing are also fixed in that last merge. I'll close this and resubmit if my compiler is still complaining when in debug. Cheers 👍

PR #1158, opened 2017-02-15
RTC - Support boot_timestamp & time_since_boot

dfct:
This PR adds boot_timestamp() & time_since_boot() to RTC, enabling us to remove the dependency on #include <OS> from the network stack classes.

OS:: & the network stack classes have been updated to call the new RTC:: functions.

### DIFF
--- a/api/kernel/os.hpp
+++ b/api/kernel/os.hpp
@@ -63,11 +63,11 @@ public:
 
   /** Timestamp for when OS was booted */
   static RTC::timestamp_t boot_timestamp()
-  { return booted_at_; }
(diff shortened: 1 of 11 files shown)


fwsGonzo:
Looks like select needs to add back OS:
```
src/posix/sys/select.cpp:136:5: error: use of undeclared identifier 'OS'
     OS::block();
```

dfct:
Yeah, looks like two posix classes were indirectly including OS from the <net> classes. Closing this & will resubmit with those classes updated (which, incidentally, mirrors the other posix classes which already had #include <kernel/os.hpp>)

PR #1159, opened 2017-02-15
RTC - Support boot_timestamp & time_since_boot

dfct:
This PR adds boot_timestamp() & time_since_boot() to RTC, enabling us to remove the dependency on #include <OS> from the network stack classes.

OS:: & the network stack classes have been updated to call the new RTC:: functions.

Additionally, two posix classes have been updated to directly #include <kernel/os.hpp>. Several already did, but these two were previously indirectly including it via the <net> classes.

### DIFF
--- a/api/kernel/os.hpp
+++ b/api/kernel/os.hpp
@@ -63,11 +63,11 @@ public:
 
   /** Timestamp for when OS was booted */
   static RTC::timestamp_t boot_timestamp()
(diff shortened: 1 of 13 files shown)


dfct:
One more missing #include <os> it looks like. I'll resubmit with an updated /test/posix/integration/udp/service.cpp

PR #1168, opened 2017-02-16
Made vmrunner into a proper python package

mnordsletten:
This removes the need for modifying the sys.path for all python scripts.

Requires installation, so I have added an installation script which is called when performing `install.sh`.

There are 3 different install "levels". Taken from the env variable: `INCLUDEOS_VMRUNNER_INSTALL`. This is mostly so that the script will not require root to run. 

1. GLOBAL - Default...Installed for all users on the system, and the ´boot´utility is put in `/usr/local/bin` and is callable from everywhere.
2. DEVELOP - Same as GLOBAL, but all changes made to vmrunner scripts are instantly updated. Good for when developing vmrunner.
3. LOCAL -Does not require root, installs for only the one user. `boot` is placed in `~/.local/bin` which will have to be added to the users `PATH`.


mnordsletten:
Now cleared up some dependency issues with the build machines. Jenkins test please

PR #1176, opened 2017-02-20
tcp: Delayed ACK, Window Scaling, Timestamps etc

AndreasAakesson:
**DACK**  (RFC 1122 p96):
More or less reduces outgoing packets with 50% when only receiving data. Tho not much of a performance gain since stopping and creating timers seems to cost just as much.
Default timeout set to `40ms` (maybe this is too low). Can be changed in application: `tcp().set_DACK(100ms);` A value of `0` will turn delayed ACK off.

**Window Scaling** (RFC 7323):
Receive and send window can now exceed 64KB, which means more data can be in flight. (Up to 1GB). Default is set to **256 KB** for trying it out. Gives a lot of performance when sending, not any when receiving (less depending on the window set..). 

**TCP Configure**:
The TCP instance that spawn new connections are now more configurable (still using default values set in `tcp/common.hpp`):
```
tcp.set_window_size(8192, 5); // 8192 bytes * 2^5 == 256 KB
// or to be used separatly
tcp.set_window_size(262144); // 256 KB, caps at 65535 without WS
tcp.set_wscale(5); // use scale 5
tcp.set_wscale(0); // don't use window scaling
tcp.set_syn_backlog(64); // allow maximum 64 outstanding connection attempts
```

**Timestamps** (RFC 7323):
Now negotiates timestamp option on handshake. Need to do some heavy refactoring to make it fully complete. Can be turned off with `tcp.set_timestamps(false);`.

### DIFF
--- a/api/net/tcp/common.hpp
+++ b/api/net/tcp/common.hpp
@@ -32,6 +32,8 @@ namespace net {
     // window scaling + window size
     static constexpr uint8_t  default_window_scaling {5};
     static constexpr uint32_t default_ws_window_size {8192 << default_window_scaling};
+    // use of timestamps option
+    static constexpr bool     default_timestamps {true};
     // maximum size of a TCP segment - later set based on MTU or peer
     static constexpr uint16_t default_mss {536};
     // the maximum amount of half-open connections per port (listener)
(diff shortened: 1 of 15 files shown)


AndreasAakesson:
jenkins test please

AndreasAakesson:
@mnordsletten :D

mnordsletten:
Should be sorted now. jenkins test please

AndreasAakesson:
jenkins test please

AndreasAakesson:
jenkins test please...

PR #1178, opened 2017-02-21
Install: Start of new install script

mnordsletten:
## Start of install script rewrite.

- Made parts of installation a lot quieter. Now only the cmake steps make noise. I'll look into suppressing output from this later.
- Fixed the function that guesses what compiler we have. It now works properly.
- When installing using terminal it now gives you a review of your install options and (y/n) prompt to continue.
- General reordering in install script and cleanup.

## When installing on a clean machine you have two options:
1. `sudo install.sh` and everything is installed in `/usr/local/includeos`
2. `install.sh` and you must set `INCLUDEOS_PREFIX`. If that is done everything is installed correctly. I created a prompt at the end urging the user to set `CC, CXX` and to add the `boot` utility to your `PATH`

### DIFF
--- a/etc/install_build_requirements.sh
+++ b/etc/install_build_requirements.sh
@@ -31,8 +31,8 @@ case $SYSTEM in
                 DEPENDENCIES="curl make clang-$clang_version nasm bridge-utils qemu jq python-jsonschema python-psutil cmake $DEPENDENCIES"
                 echo ">>> Installing dependencies (requires sudo):"
                 echo "    Packages: $DEPENDENCIES"
-                sudo apt-get -q update || exit 1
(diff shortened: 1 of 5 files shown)


mnordsletten:
This can be merged in it's current state, but there is more work to do.

1. If you call the install script without `sudo` and no `INCLUDEOS_PREFIX` it will most likely fail due to not having access to `/usr/local/`. A test for this before beginning the installation is needed.
2. The dependency situation is still the same as before. I'm planning to look at this next. The plan is to pull as much platform specific stuff into the same place where it can be easily managed.

mnordsletten:
Jenkins test please

PR #1187, opened 2017-02-24
Delegate empty check fix

Voultapher:
Not happy about it :|
Still stability is better than minor optimization.


RicoAntonioFelix:
Wisdom state the following:
(slow and correct) > (fast and incorrect)
:wink:

PR #1197, opened 2017-02-28
kernel: Augment memmap module

RicoAntonioFelix:
# Change
* Add API documentation
* Add `const noexcept` to methods that exhibit such behaviour
* Move implementation of methods that contain more than one statement to the implementation file


fwsGonzo:
Is there any quick way you can change the comments from /// to using stars `/** **/`?

PR #1220, opened 2017-03-06
net: Augment socket module

RicoAntonioFelix:
# Change
* Add API documentation
* Add `const noexcept` to methods that exhibit such behaviour


fwsGonzo:
Might have to consider byte ordering when you do the less than comparison on Socket

does it work as intended now?

RicoAntonioFelix:
I presume it does work as intended :smile:

PR #1243, opened 2017-03-16
net: Add documentation to ip4::Header

RicoAntonioFelix:
# Change
* Add documentation to the fields in the IPv4 header


alfred-bratterud:
👍 I've rewritten the TOS-field due to newer RFC's but apart from that your doc is better :-)

RicoAntonioFelix:
Okay :smile:

PR #1246, opened 2017-03-20
examples: tweaks to demo_service

ingve:


alfred-bratterud:
@ingve this solution is necessary to get the example working on GCE so I'll merge - but it makes the  service slower to start if you don't have a DHCP server running. Could you add a printf telling the user why we're waiting? e.g. `waiting 5 sec. for DHCP, falling back to static config` or something

PR #1251, opened 2017-03-21
net: Refactored module ip4::Addr

RicoAntonioFelix:
# Change
* Made ip4::Addr a literal type


fwsGonzo:
Are you sure this compiles on G++? There is an open issue regarding the effort of just compiling the OS with g++ (without changing anything else).

RicoAntonioFelix:
ip4

Compiled for `g++ version 6.2.0`

PR #1255, opened 2017-03-22
Update HTTP API header file

RicoAntonioFelix:


alfred-bratterud:
Totally agree - would also be nice if the demo service used the server

PR #1256, opened 2017-03-22
Update HTTPS API header file

RicoAntonioFelix:


AndreasAakesson:
jenkins test please

api/https
```diff
@@ -17,11 +17,10 @@
...
 
 #pragma once
-#ifndef API_HTTPS_HEADER
-#define API_HTTPS_HEADER
+#ifndef INCLUDEOS_HTTPS_API_HPP
+#define INCLUDEOS_HTTPS_API_HPP
 
-#include "net/http/response.hpp"
-#include "net/http/request.hpp"
+#include "net/http/client.hpp"
```
AndreasAakesson: We currently don't have a TLS Client (only Server), so I'm not sure it's "correct" to include the client here. Thoughts?

RicoAntonioFelix: You're right :smile:

PR #1261, opened 2017-03-23
Improvements to bundle creation

mnordsletten:
`etc/create_source_bundle.sh` is now the one stop shop for creating bundles. 

Also added option to `install.sh` to install from a local bundle. Use `install.sh -b <path_to_bundle>` to use this.


mnordsletten:
The tests that interface with Openstack need additional exit codes that specify errors on openstacks end. Those tests can then be rerun. Jenkins test please

mnordsletten:
I've now verified manually that the openstack tests pass. Jenkins test please

PR #1262, opened 2017-03-23
Make CPU features optional

AndreasAakesson:
Since AVX was introduced there was some issues with running on Mac with qemu (it didnt work).

This PR makes vmrunner have a default config that's installed in `includeos/vmrunner/vm.default.json`. This can be edited to your own liking. Additional local `vm.json` will extend and override any duplicate entry (it has higher priority).

I also had to add an CMake option to build IncludeOS different (without AVX etc), and called this `cpu_feat_vanilla` which is by default **ON**. This also determine which default vm config that should be installed.


AndreasAakesson:
jenkins test please

AndreasAakesson:
jenkins test please

PR #1265, opened 2017-03-24
Make CPU features optional

AndreasAakesson:
See #1262

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,16 +23,22 @@ execute_process(COMMAND git describe --dirty
 	        OUTPUT_VARIABLE OS_VERSION)
 string(STRIP ${OS_VERSION} OS_VERSION)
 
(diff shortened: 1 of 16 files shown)


AndreasAakesson:
So this turned out really ugly. Opinions about naming or how to do it is appreciated. Need a solution for `start.asm`

mnordsletten:
The hping fail seems to be from two tests trying to access some resource at the same time. Highly random, so I will have to prevent this from happening. Jenkins test please

mnordsletten:
This should be fixed now. There was an apt-get process running on the system when starting the test, thus failing all apt-get processes we started. It should now wait for the apt-get resource to be free.
Jenkins test please

mnordsletten:
I wont make any promises anymore, but let's give it a go. Jenkins test please

PR #1278, opened 2017-03-30
unit_tests: added --pass to misc unit test run, to see passed tests

mnordsletten:


mnordsletten:
@alfred-bratterud I'm getting an encoding error with the test output from the unittests. I think this unit test is generating some sort of bad output, could you take a look:
```
/home/ubuntu/IncludeOS/test/util/unit/chunk_test.cpp:40:
passed: Constructing a chunk with a shared buffer and size:
c for { 'h', ' ', 'T', '�', 'h', ' ', 'T', '�', '�',
'�', '�', '', '�', '�', '�', '', '
```

alfred-bratterud:
jenkins test please. @AndreasAakesson any comments?

AndreasAakesson:
Yes, f74f3c9 solves the problem. Tho only the one in chunk, there is probably more in other unit tests. Example is `char* == char*` which will probably also print bs.

mnordsletten:
After fixing that one instance I also noticed that this was happening again and again, so instead I set out to fix this problem in python and try to handle it there. Seems like I didn't quite fix it all the way though. 

On the plus side I now know a lot more about encodings and unicode in python.

AndreasAakesson:
How can I replicate the cases where it fails? The other solution is that you gather all the cases where it spews garbage and I can fix those cases. Or just do it with python :D

PR #1222, opened 2017-03-06
Statman and ICMP

AnnikaH:

PR #1280, opened 2017-03-30
Updates based on feedback on #1222 and more

AnnikaH:
- Updates based on feedback on pull request #1222 
- Socket moved out of namespace tcp
- Error handling with net::Error (and net::ICMP_error)
- Handling of Error callbacks in UDP
- DNS integration test
and more


api/net/ip4/udp.hpp
```diff
@@ -141,27 +150,52 @@ namespace net {
...
-    port_t      current_port_;
+    port_t                      current_port_;
 
     // the async send queue
     std::deque<WriteBuffer> sendq;
+
+    /** Error entries are just error callbacks and timestamps */
+    class Error_entry {
+    public:
+      Error_entry(UDP::error_handler cb) noexcept
```
AndreasAakesson: By moving the implementation of the constructor to the cpp you can avoid including `<rtc>` in the header (reducing the scope).

api/net/ip4/udp.hpp
```diff
@@ -141,27 +150,52 @@ namespace net {
...
+
+      UDP::error_handler callback;
+
+    private:
+      RTC::timestamp_t timestamp;
+
+    }; //< class Error_entry
+
+    /** The error callbacks that the user has sent in via the UDPSockets' sendto and bcast methods */
+    std::unordered_map<Socket, Error_entry, Socket::pair_hash> error_callbacks_;
```
AndreasAakesson: A implementation of `std::hash<Socket>` would make it possible to drop the third template arg `Socket::pair_hash` (this function can be called inside the std::hash`). See http://en.cppreference.com/w/cpp/utility/hash

api/net/socket.hpp
```diff
@@ -138,12 +147,37 @@ class Socket {
...
 
-} //< namespace tcp
+class Quadruple {
+
+public:
+  Quadruple() noexcept
+    : source_{}, destination_{}
+  {}
+
+  Quadruple(const Socket::Address src_address, const Socket::port_t src_port,
```
AndreasAakesson: Maybe add additional constructor `Quadruple(Socket, Socket)`.

api/net/ip4/udp.hpp
AnnikaH: 👍

api/net/socket.hpp
AnnikaH: 👍

api/net/inet.hpp
```diff
@@ -44,28 +44,28 @@ namespace net {
     using Route_checker = delegate<bool(typename IPV::addr)>;
     using IP_packet_factory = delegate<typename IPV::IP_packet_ptr(Protocol)>;
 
-    using Error_type = icmp4::Type;
-    using Error_code = uint8_t;
-
     template <typename IPv>
-    using resolve_func = delegate<void(typename IPv::addr)>;
+    using resolve_func = delegate<void(typename IPv::addr, Error&)>;
```
AnnikaH: No seems like Error& can be const here =)

api/net/ip4/udp.hpp
AnnikaH: private attribute RTC::timestamp_t, but could consider int64_t

alfred-bratterud:
jenkins test please

mnordsletten:
This is the error I was seeing yesterday as well @AnnikaH where it seems like you were not expecting the 10.0.0.1 dns server to actually give a valid response. (but it seems like it does)

`Resolved IP address of hotmail.com with DNS server 10.0.0.1: 157.56.198.220`

PR #1287, opened 2017-03-31
vmware: add some error checking, support for launching multiple instances

ingve:


fwsGonzo:
The failure (else) cases are just ignored right now.

ingve:
No, if you run the script now you will see that the output from the failed commands is displayed, telling the user exactly what went wrong, but the script doesn't try to continue once something has gone wrong.

alfred-bratterud:
jenkins test please

PR #1291, opened 2017-04-04
TCP and CMake

AndreasAakesson:
Small error in TCP.

Also made it possible to easier include IncludeOS libs.

```
# THIRD PARTY LIBRARIES:

set(LIBRARIES
  "libmana.a"
  )
```

This will first search for `libmana.a` locally. If not found, it tries the `includeos/lib` directory. If nothing there it fails as usual.


AndreasAakesson:
jenkins test please

AndreasAakesson:
Also fixed a IP and HTTP client issue.

PR #1301, opened 2017-04-11
Path MTU Discovery

AnnikaH:
Also some unit tests and const Error& where possible (previously not const)
Note: Path MTU Discovery is enabled by default and has not been tested thoroughly (hard to find a router that responds with an ICMP Too Big message (ICMP error message of type Destination Unreachable and code Fragmentation Needed))


ricarkol:
Just wanted to mention some numbers regarding ukvm. Increasing the MTU from 1500 to 9000 increases TCP (send) from 670 Mbits/sec to 4.69 Gbits/sec. Also, it's currently hardcoded here: https://github.com/hioa-cs/IncludeOS/blob/dev/src/drivers/solo5net.hpp#L52 (that's what i changed for the tests).

PR #1312, opened 2017-04-26
WIP: pthreads-compatible fibers

alfred-bratterud:
This is very much a PoC, but so far it looks like we can have pthreads-compatible fibers, while providing a more type-safe C++ API.  Fibers now support a single parameter with void* storage, and return value, also with void* storage.


api/kernel/fiber.hpp
```diff
@@ -19,45 +19,134 @@
...
   using Stack_ptr = std::unique_ptr<char[]>;
   static constexpr int default_stack_size = 0x10000;
   static constexpr int redzone = 0;
 
   void start();
   void restart();
 
-  Fiber(Fiber* parent, int stack_size, init_func func)
+  template<typename R, typename P>
+  Fiber(Fiber* parent, int stack_size, R(*func)(P), void* arg)
```
RicoAntonioFelix: What's the valid set of values for `stack_size`?
Is it [..., -2, -1, 0, 1, 2, ...]?

fwsGonzo: it's stack size in bytes, and should be some value in pages (or at least cache-aligned), so the smallest stack is 64b, while realistic minimum is 4kb. On a normal OS (and for IncludeOS) the main stack is 1mb

RicoAntonioFelix: Okay, I understand the rationale now... :+1:
Should probably change the type from `int` to `size_t` to constrain the valid values to natural numbers...

PR #1320, opened 2017-05-05
platforms: build for different device models

alfred-bratterud:
Where "arch" specifies the CPU architecture, the "platform" specifies a device model and decides what the OS does before starting the service.  The x86_pc platform is a full blown computer, with PCI devices, APIC timers etc., whereas "x86_nano" only has a serial port and possibly timers.


fwsGonzo:
From looking at the nano code, it should be possible right now to not use APIC or ACPI, by implementing a few stubs (eg. make arch legacy IRQ stuff no-ops)

PR #1327, opened 2017-05-09
bash: Now uses proper syntax for setting variables

mnordsletten:
Mostly a test to see if new jenkins triggers correctly.


mnordsletten:
Jenkins test please

PR #1334, opened 2017-05-13
Changes to Inet on_config

AndreasAakesson:
You can now ask Inet if its configured, or setup one or more on_config events. These will only be called when the stack is being configured (no timeout). Negotiating DHCP works as usual (with the timeout callback). DHCP no longer got a vector of config handler but one on_config (default is printing to console that it has timed out).

Opinions?


AndreasAakesson:
jenkins test please

PR #1366, opened 2017-05-22
SMP now stable in x86_64

fwsGonzo:
Statman was too small with only 2 pages, and the all the cores couldn't share the same stack


mnordsletten:
This did not trigger due to `Couldn't obtain mergeable status.` I will try and trigger this test again.

Jenkins test please

PR #1382, opened 2017-05-31
C++ Exceptions working in SMP

fwsGonzo:
A new bundle has to be created with thread-safe ABI glue
- Conditionals need to be implemented properly, as services won't always start properly without it
- Mutexes can probably just be spinlocks for now, especially because the SMP systems are made to be 1-to-1, until someone needs multi-threading


alfred-bratterud:
Right, so building libc++abi / libc++ with threads and providing mutex will make exceptions thread safe. Any idea what the overhead cost of this is for a non-threaded application?

fwsGonzo:
Not worried about performance. Print when mutexes are locked or conditionals broadcast or waited on to see.

PR #1386, opened 2017-06-07
Removed g++-multilib from fedora dependencies

oyviaase:
The g++-multilib package is not in the fedora repositories. 

This causes the installation to crash when it tries to install the dependencies.

### DIFF
--- a/etc/install_dependencies_linux.sh
+++ b/etc/install_dependencies_linux.sh
@@ -141,7 +141,8 @@ case $SYSTEM in
                 sudo apt-get -qqy install $DEPENDENCIES > /dev/null || exit 1
                 ;;
             "fedora")
-                DEPENDENCIES="$DEPENDENCIES"
+                # Removes g++-multilib from dependencies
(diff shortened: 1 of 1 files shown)


ingve:
jenkins test please

alfred-bratterud:
Hi @oyviaase thanks for contributing! Afaik. the multilib is needed to build any 32-bit parts with GCC. Currently I think we only do that when building the 32-bit dependencies, such as newlib for 32-bit etc. Could you try to run [create_binary_bundle.sh](https://github.com/hioa-cs/IncludeOS/blob/master/etc/create_binary_bundle.sh) and verify that this works? It's clearly more important that a normal install works on fedora, than being able to build everything from source, but if this is just an issue with e.g. package naming it would be nice to identify the root cause and keep as much support as possible.

oyviaase:
The script still works. My guess is that fedora includes the 32-bits parts  in the gcc packages for x86_64, insteadof seperating it into another package. Since a search for multilib with dnf provides 4 packages
- multilib-rpm-config
- python2-multilib
- python3-multilib
- python-multilib-conf

PR #1389, opened 2017-06-13
Install: option to disable bridge install

mnordsletten:
Useful for installation with docker


mnordsletten:
Failed due to it not being able to download botan from github. Jenkins test please

Issue #1388, opened 2017-06-09
[BUG] Build error on latest dev branch


RicoAntonioFelix:
The latest `dev` branch contain the following error while building the `OS`:
```
/home/rico/IncludeOS/src/kernel/irq_manager.cpp:136:8: error: %rip can only be
      used as a base register
  asm ("movq %%rip, %0" : "=d" (regs[16]));
       ^
<inline asm>:1:7: note: instantiated into assembly here
        movq %rip, %rdx
             ^~~~~
```

mnordsletten:
I'm not able to reproduce this on a fresh install of `ubuntu 16.04`. Any more information you can provide?

RicoAntonioFelix:
Hey @mnordsletten, the platform I'm currently using has the following specs:
### OS Information
```
PRETTY_NAME="Parrot GNU/Linux 3.6 JollyRoger"
NAME="Parrot GNU/Linux"
ID=parrot
VERSION="3.6 (JollyRoger)"
VERSION_ID="3.6"
ID_LIKE=debian
ANSI_COLOR="1;31"
HOME_URL="https://www.parrotsec.org/"
SUPPORT_URL="https://community.parrotsec.org/"
BUG_REPORT_URL="https://dev.parrotsec.org/"
```
### Compiler Information
```
clang version 4.0.0 (tags/RELEASE_400/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/clang/bin
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/5
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/5.4.1
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/6
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/6.3.0
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/6.3.0
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Candidate multilib: x32;@mx32
Selected multilib: .;@m64
```

mnordsletten:
Thanks. 
I can confirm that the error is happening on clang-4.0 for me as well. We will look into this.

PR #1405, opened 2017-07-07
kernel: Fix register dump of the instruction pointer on x86_64

mmha:
This should address #1388. I am able to compile IncludeOS using the current clang SVN.

See https://stackoverflow.com/questions/12397451/why-cant-i-save-the-value-of-rip for an explanation.


ingve:
ok to test

ingve:
Awesome! Thank you for contributing!

PR #1406, opened 2017-07-12
Fix IPv4 address validation

krzysztofwos:
The regex had an en-dash instead of a hyphen in the character range

### DIFF
--- a/api/net/ip4/addr.hpp
+++ b/api/net/ip4/addr.hpp
@@ -107,10 +107,9 @@ struct Addr {
 
     const static std::regex ipv4_address_pattern
     {
-      "^\\s*(25[0–5]|2[0–4]\\d|[01]?\\d\\d?)\\."
-        "(25[0–5]|2[0–4]\\d|[01]?\\d\\d?)\\."
-        "(25[0–5]|2[0–4]\\d|[01]?\\d\\d?)\\."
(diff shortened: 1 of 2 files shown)


ingve:
ok to test

api/net/ip4/addr.hpp
```diff
@@ -107,10 +107,9 @@ struct Addr {
 
     const static std::regex ipv4_address_pattern
     {
-      "^\\s*(25[0–5]|2[0–4]\\d|[01]?\\d\\d?)\\."
-        "(25[0–5]|2[0–4]\\d|[01]?\\d\\d?)\\."
-        "(25[0–5]|2[0–4]\\d|[01]?\\d\\d?)\\."
-        "(25[0–5]|2[0–4]\\d|[01]?\\d\\d?)\\s*$"
+#define OCTET_PATTERN "(25[0-5]|2[0-4]\\d|[01]?\\d\\d?)"
+      "^\\s*" OCTET_PATTERN "\\." OCTET_PATTERN "\\." OCTET_PATTERN "\\." OCTET_PATTERN "\\s*$",
```
RicoAntonioFelix: Is that a trailing comma(**,**) I'm seeing there?

Although it's calling the second constructor on [this](http://en.cppreference.com/w/cpp/regex/basic_regex/basic_regex) list, it's better off not being there...

krzysztofwos: Fixed

PR #1413, opened 2017-08-06
Code Tweak[delegate.hpp]

FallingFromBed:


ingve:
jenkins test please

FallingFromBed:
Ooops, The pack_first_t I've modified worked for me https://godbolt.org/g/ntNH9T .looking on why this error.
```
 il/delegate.hpp:494:37: error: pack expansion used as argument for non-pack parameter of alias template
20:01:56                 std::is_same<detail::pack_first_t<Args...>, C>::value, int> = 0
```

PR #1443, opened 2017-08-25
IP4 addr operator overloading

AnnikaH:
- Bug fix operator >
- Added >= and <=

### DIFF
--- a/api/net/ip4/addr.hpp
+++ b/api/net/ip4/addr.hpp
@@ -217,6 +217,30 @@ struct Addr {
   { return ntohl(whole) < ntohl(raw_addr); }
 
   /**
+   * Operator to check for less-than-or-equal relationship
+   *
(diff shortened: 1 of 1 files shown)


fwsGonzo:
Can you clone dev from scratch and then create this change as a single commit?  It would reduce the number of commits greatly. 👍

fwsGonzo:
Your commit named "IP4 addr operator overloading" has 14,000 changes now :P

AnnikaH:
Hm yes, I used git reset --soft to remove all the Merge branch-commits, but then when committing, the earlier versions of the files where noted as changes to the branch, so I had to merge branch dev to get the latest. I can try to reset --soft again, but pull from dev before committing - hopfully that will work.

AnnikaH:
Think that did the trick

PR #1461, opened 2017-09-01
lib: Make LiveUpdate::resume boolean

fwsGonzo:
Also rename `register_serialization_callback` to simply `register_partition`.

Resume setup is simplified if normal execution simply returns false if there is nothing to resume.


fwsGonzo:
jenkins test please

PR #1464, opened 2017-09-04
solo5: Force page-aligment for first ELF segment

ricarkol:
Force the ELF's first segment to be page aligned.

From this (from `readelf -a`):
```
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x00000000000000e8 0x0000000000a000e8 0x0000000000a000e8 <===not aligned
                 0x00000000002c3ecb 0x0000000000455e28  RWE    80
  GNU_EH_FRAME   0x0000000000204138 0x0000000000c04138 0x0000000000c04138
                 0x000000000000873c 0x000000000000873c  R      4
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RWE    10
```
to this:
```
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000000100 0x0000000000a01000 0x0000000000a01000
                 0x00000000002c3eb3 0x0000000000455f10  RWE    80
  GNU_EH_FRAME   0x0000000000204138 0x0000000000c05038 0x0000000000c05038
                 0x000000000000873c 0x000000000000873c  R      4
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RWE    10
```

### DIFF
--- a/src/arch/x86_64/linker.ld
+++ b/src/arch/x86_64/linker.ld
@@ -23,7 +23,8 @@ SECTIONS
   PROVIDE ( _ELF_START_ = . + 0xA00000);
   PROVIDE ( _LOAD_START_ = _ELF_START_); /* For convenience w. multiboot */
 
-  .multiboot (_ELF_START_ + SIZEOF_HEADERS): {
+  . = _ELF_START_ + SIZEOF_HEADERS;
+  .multiboot ALIGN(0x1000): {
       PROVIDE(_MULTIBOOT_START_ = .);
       *(.multiboot)
    }


ingve:
add to whitelist

ricarkol:
I'm getting the same errors on every unikernel I tried even without this change (am I missing something?):

```
Starting VM with Qemu. 
(Once inside Qemu, 'Ctrl+a c' will enter the Qemu console, from which you can type 'q' to quit.)

You should now get a boot message from the virtual machine:
[ WARNING ] WARNING: Environment variable INCLUDEOS_PREFIX is not set. Trying default /usr/local
Looking for chainloader: 
Found /usr/local/includeos/chainloader Type:  ELF 32-bit LSB executable, Intel 80386, version 1 (SYSV), statically linked, stripped
[ WARNING ] Running with sudo

>>>> !!! CPU 0 EXCEPTION !!! <<<<
    Invalid Opcode (6)   EIP  0xef000   CODE 0

  EAX:  00002000  EBP:  0009fb58
  EBX:  002a5434  ESP:  0000200c
  ECX:  00009500  ESI:  0062c3a4
  EDX:  2badb002  EDI:  00676fec

  EIP:  000ef000  EFL:  00000006


	**** PANIC ****
CPU: 0, Reason: Invalid Opcode (6)
Heap is at: 0x678000 / 0xfffffff  (diff=261652479)
Heap usage: 4 / 255523 Kb

[0] 0x00104360 + 0x103: panic
[1] 0x001037d0 + 0x0be: __cpu_exception
[2] 0x00103abd + 0x000: __cpu_except_7
[3] 0x000ef000 + 0x000: 0xef000
[4] 0x001002b4 + 0x132: Service::start()
[5] 0x001026b0 + 0x067: kernel_start
[6] 0x001039a0 + 0x01f: __arch_start
[7] 0x001024ae + 0x03d: rock_bottom
```

alfred-bratterud:
@ricarkol is it possible that aligning the multiboot header causes problems for the bootlader? (Qemu `-kernel` in this case)
> The Multiboot header must be contained completely within the first 8192 bytes of the OS image, and must be longword (32-bit) aligned. In general, it should come as early as possible, and may be embedded in the beginning of the text segment after the real executable header.

It's a bit vague [in the spec.](https://www.gnu.org/software/grub/manual/multiboot/multiboot.html#OS-image-format) but it should come "as early as possible". I'll try to take a look later today and see if I can reproduce.

fwsGonzo:
It's already page-aligned, just at the ELF header and not at .multiboot. The kernel is loaded at a page-aligned area, which can seem offset if its loaded with an ELF loader. So, the ELF loader just has to recognize that `multiboot - ELF_HEADERS == kernel base (aligned)`.

Not to mention that it doesn't matter that it's not aligned as we're talking about 100+ bytes of offset from a page-aligned area which is still deemed part of the kernel area. So the ELF loader could be doing `first_section & 0xFFFFF000`.

alfred-bratterud:
@ricarkol I believe the issue is that our chainloader / legacy bootloader expects the whole ELF binary to reside in memory when we boot, not only the individual sections. When the chainloader / legacy bootloader copies the ELF it copies the whole thing, including the ELF header. To account for this we add the size of the ELF header to the start of the binary in the linker script (`+ SIZEOF_HEADERS`), which in turn skews the load address out of alignment. For the chainloader the fix is easy enough - we can just properly load the segments instead of the whole ELF (tired that now and it works). That's a bit trickier for the legacy bootloader however, as it can only hold 512 bytes of code. Then again, I don't think we use that for anything so it might be time to say goodbye. That being said - GRUB loads our image just fine (and yours - try with `boot -g .`) so it doesn't seem to have the same alignment requirement ukvm has. @fwsGonzo any thoughts?

alfred-bratterud:
@ricarkol we had a discussion about this and we'll probably align the first section and update chainloader / bootloader accordingly.

ricarkol:
Thanks @alfred-bratterud 

Regarding "what reason's you'd have to force page alignment of a section". ukvm could load the ELF, actually it does, but what it can't do is enforcing memory protection of non-aligned ELF segments. The reason is that the syscall used for this is `mprotect`, and it only works with page aligned pieces of memory. It makes sense actually, as it will be using page tables on the host to do so.

PR #1478, opened 2017-09-13
Point to upstream Solo5

ricarkol:
Two changes:
- Point to solo5 master as of today, commit `f8a277f83807`.
- Don't rely on `make build` (which doesn't exist in master), and just copy solo5.o, ukvm-bin, and solo5.h by hand.

Tested with `test_ukvm.sh` on a docker container (started with `--privileged --net=host`) running Ubuntu 16.04.3 LTS and gcc 5.4.0.

cc @alfred-bratterud


alfred-bratterud:
Great @ricarkol thanks!

PR #1479, opened 2017-09-13
Fix initialization problems with high BSS value

fwsGonzo:
Also move kernel_start to platform as it was already halfway there


fwsGonzo:
jenkins test please

PR #1487, opened 2017-09-21
net: Moved Filter_chain into IP (fix #1486)

### LINKED ISSUE #1486: Filter_chain definitions prevents building with g++
A Filter_chain alias that refers to its own name, which is templated, is invalid in g++. Solved when g++ build succeeds.
```
.../includeos/api/net/inet.hpp:128:43: error: declaration of 'using Filter_chain = struct net::Filter_chain<IPV>' [-fpermissive]
     using Filter_chain = Filter_chain<IPV>;
```

AndreasAakesson:


AndreasAakesson:
I also removed forward chain (for now). I don't think it should be located in IP but maybe the Router?

PR #1494, opened 2017-09-26
Add support to boot for connecting the vm to an arbitrary bridge

perbu:
I've modified the boot script to attach VMs to arbitrary VM. @AndreasAakesson mentioned he needs this in order to simplify the injection of VLAN-tagged packets.

I've also supplied a qemu-ifdown so qemu doesn't whine as much.


perbu:
The second commit shouldn't have gotten in there (branches got tangled up). Mostly it is there to make demos look cooler as you don't have to spend time making a cmakelist.txt-file.

AndreasAakesson:
jenkins test please

PR #1497, opened 2017-09-28
IP4 Cidr class

AnnikaH:
Suggestion

### DIFF
--- a/api/net/ip4/cidr.hpp
+++ b/api/net/ip4/cidr.hpp
@@ -29,6 +29,24 @@ public:
   /**
    * @brief      Constructor
    *
+   * Create an IPv4 cidr object to represent the cidr <address/mask>,
+   * f.ex. 10.0.0.0/24
+   *
+   * @param[in]  address  IPv4 address
+   * @param[in]  mask     A number between 0 and 32, representing the number
+   *                        of leading 1 bits in the netmask, f.ex.:
+   *                        32 represents the netmask 255.255.255.255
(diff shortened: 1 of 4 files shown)


api/net/ip4/cidr.hpp
```diff
@@ -0,0 +1,65 @@
...
+#define NET_IP4_CIDR_HPP
+
+#include "addr.hpp"
+
+namespace net {
+namespace ip4 {
+
+class Cidr {
+public:
+  Cidr(uint8_t p1, uint8_t p2, uint8_t p3, uint8_t p4, uint8_t mask)
```
AndreasAakesson: * Can be probably be `constexpr` (if members are made `const`, which they can since the class is immutable), and `noexcept`.
* Not sure if any need for Cidr(Addr&& addr, uint8_t mask) (maybe just wait with it?)
* Some short explanation of how to use the constructor would be nice!

api/net/ip4/cidr.hpp
```diff
@@ -0,0 +1,65 @@
...
+    from_ = Addr{from_ip};
+    to_ = Addr{to_ip};
+  }
+
+  // Possibly not necessary
+  Cidr(Addr from, Addr to)
+  : from_{from}, to_{to}
+  {}
+
+  bool contains(Addr ip) {
```
AndreasAakesson: This and the following functions could have `const noexcept`.

PR #1501, opened 2017-10-02
VLAN support

AndreasAakesson:
A start for VLAN support. This forced some changes to `Nic`, which also showed that things gets a bit complicated below IP.

Don't merge yet, would like some feedback and maybe one more iteration. Also need to add support for the "transmit queue available" event (need to support multiple on the physical Nic for broadcasting to the virtual once).

By measuring the heap increase before and after the VLAN-stack is created I got about a 65kb increase, which for me sounds ok.


AndreasAakesson:
jenkins test please

AndreasAakesson:
jenkins test please

PR #1517, opened 2017-10-10
VLAN integration test

AndreasAakesson:
Just a really simple one.

Do not work with vmxnet3 in qemu for some reason.


mnordsletten:
Let us see if we can get Jenkins to test again. Jenkins test please

mnordsletten:
Let us see if we can get Jenkins to test again. Jenkins test please

mnordsletten:
Sorry for spam. We need to make sure this is working. Jenkins test please

PR #1530, opened 2017-10-18
Auto create CMakelist from seed if not present

perbu:
Allows for very nice demos by just creating service.cpp. cmake config is then copied from seed.


ingve:
add to whitelist

PR #1534, opened 2017-10-21
Adds vm.json file for snake example

SeanMooney:
- IncludeOS now uses external vm config.
- The snake example requires graphics output
  to function correctly.
- This change defines a new vm.json file
  which enables qxl graphics and 32M of ram.


ingve:
ok to test

PR #1536, opened 2017-10-23
Add support for arbitrary bridge names

perbu:
Currently we have hardcoded the use of bridge43. This pull enables vmrunner to attach the vm to an arbitrary bridge on the system. Useful if you want the VM to enter a specific bridge (perhaps one used by vmware). Useful for testing.

It disables cleaning of the environment done by sudo when starting qemu. I cannot see any risks here.


perbu:
I'm obviously missing something here since this seems to break on Linux. Seem to work at my system, though. :-/

Issue #1540, opened 2017-10-25
error: incompatible implicit declaration of built-in function malloc/calloc


mellowiz:
Hi all,
./install.sh would halt while highlighting the error in the subject.
I've found out that the only way to move further is by adding "#include <stdlib.h>" to both "ukvm_hv_kvm.c" and "ukvm_hv_kvm_x86_64.c" and recompile.

Is anyone able to comment on this?

OS: Lubuntu 17.10 4.13.0-16-generic x86_64

Thanks,
Mel

mellowiz:
Actual error message:
...
/home/<username>/github/IncludeOS/build_x86_64/precompiled/src/solo5_repo/ukvm/ukvm_hv_kvm.c: In function ‘ukvm_hv_init’:
/home/<username>/github/IncludeOS/build_x86_64/precompiled/src/solo5_repo/ukvm/ukvm_hv_kvm.c:42:26: error: implicit declaration of function ‘malloc’ [-Werror=implicit-function-declaration]
     struct ukvm_hv *hv = malloc(sizeof (struct ukvm_hv));
                          ^~~~~~
/home/<username>/github/IncludeOS/build_x86_64/precompiled/src/solo5_repo/ukvm/ukvm_hv_kvm.c:42:26: error: incompatible implicit declaration of built-in function ‘malloc’ [-Werror]
/home/<username>/github/IncludeOS/build_x86_64/precompiled/src/solo5_repo/ukvm/ukvm_hv_kvm.c:42:26: note: include ‘<stdlib.h>’ or provide a declaration of ‘malloc’
...

HTH,
Mel

fwsGonzo:
I have the same error and it's a problem with compiling SOLO5. For now you can just go into 64-bit build folder (build_x86_64) and do `ccmake ..`, then disable building Solo5.

Or, maybe its easier to just go into each of those .c files and `#include <malloc.h>`

ricarkol:
Created #1545 to point to latest solo5 (which fixes this).

fwsGonzo:
https://github.com/hioa-cs/IncludeOS/pull/1545 which is merged now fixed it :+1:

PR #1545, opened 2017-10-25
solo5: point to solo5 latest

ricarkol:
Point to latest solo5, so to get the fix for #1540 . Just in case, the added `include`s (in latest master) are here:
https://github.com/Solo5/solo5/blob/master/ukvm/ukvm_hv_kvm.c#L32
https://github.com/Solo5/solo5/blob/master/ukvm/ukvm_hv_kvm_x86_64.c#L34

Should I create a PR for master as well?

@fwsGonzo


ingve:
add to whitelist

PR #1551, opened 2017-10-31
New libs

alfred-bratterud:
Update newlib, libc++, gcc, binutils to latest. Also updates clang to 4.0 - it's tested to work with 5.0, but that's not available on current ubuntu LTS (16.04)


src/posix/pthread.cpp
```diff
@@ -68,27 +68,27 @@ int pthread_equal(pthread_t t1, pthread_t t2)
...
   //SMP_PRINT("Spin locked %p\n", mutex);
   return 0;
 }
 int pthread_mutex_trylock(pthread_mutex_t *mutex)
 {
   if (mutex == nullptr) return 1;
-  if (mutex->spinlock) return 1;
-  lock(mutex->spinlock);
+  if (mutex) return 1;
+  lock(*mutex);
```
RicoAntonioFelix: This statement will never be reached due to the two statements above it...

PR #1560, opened 2017-11-09
linux: Update linux platform and add support for thinLTO

fwsGonzo:
This makes the Linux platform buildable again on dev


fwsGonzo:
jenkins test please

Issue #1521, opened 2017-10-13
Solo5/uvkm: automated test


alfred-bratterud:
Add an automated test for Solo5 integration. E.g. `./test/misc/integration/ukvm` that starts the demo service with ukvm and verifies ping / curl response.

niks3089:
Is this already done? I wanted to pick this up but I see that test_ukvm.sh already tests demo_service unikernel

ricarkol:
Yes @niks3089 this should be done.

@alfred-bratterud: are you OK with closing this?

fwsGonzo:
It's being run as part of `misc` group on jenkins automatically, so I closed this

PR #1575, opened 2017-11-16
solo5: add missing __arch_time_now to solo5 platform

ricarkol:
Adding missing `__arch_time_now` implementation to solo5. This was broken after 5b90cfc277c3284c9d2b6c65f605e5bc7282eff9. Will work on something for #1521 

Question: the returned time should be in nanoseconds right?

Tested with test_ukvm.sh on an ubuntu:16.04 container.

### DIFF
--- a/src/platform/x86_solo5/os.cpp
+++ b/src/platform/x86_solo5/os.cpp
@@ -43,6 +43,12 @@ void solo5_poweroff()
   for(;;);
 }
 
+// returns wall clock time in nanoseconds since the UNIX epoch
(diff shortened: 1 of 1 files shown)


fwsGonzo:
its the wallclock time, but theres no reason we cant just decide its in nanos, since it should fit nicely in 64-bits, right?

ricarkol:
Thanks @fwsGonzo , changed the call to `solo5_clock_wall`, which returns the wall clock time in nanoseconds since the UNIX epoch.

PR #1582, opened 2017-11-20
(WIP) solo5: adding misc/integration/ukvm/test.sh

ricarkol:
This is a new solo5 test along the lines of #1521 : "./test/misc/integration/ukvm that starts the demo service with ukvm and verifies ping / curl response".

It returns success (rc=0) and failure (rc=1) in the following cases:
1. if non-linux, it returns success (rc=0) and prints "SKIPPED".
2. linux without access to KVM returns success (rc=0) and prints "SKIPPED" if the unikernel can be built.
3. linux without access to KVM returns failure (rc=1) and prints "FAILURE" if the unikernel can NOT be built.
4. linux with access to KVM returns success and prints "SUCCESS" if the unikernel can be built, executed, and curl'ed correctly.
5. linux with access to KVM returns failure and prints "FAILURE" if the unikernel can NOT be built, or it can NOT be executed, or it can NOT be curl'ed within 30 seconds.

The test also has a '-v' argument that can be used to print all logs (stdout/stderr from the unikernel and the curl command). In case of failure, all logs are printed.

Given that ukvm needs access to KVM, I don't think it can run in jenkins or travis, but I'm not sure. At the very least, this test checks that the demo_service unikernel can be built correctly (2 or 3 above). So, my question for you is what to do next.

### DIFF
--- a/test/misc/ukvm/test.sh
+++ b/test/misc/ukvm/test.sh
@@ -1,0 +1,177 @@
+#!/bin/bash
+
+# Runs the IncludeOS demo service, and tests it by doing a curl.
+
+NAME=demo_service
+NET_IP=10.0.0.42
+NET_DEVICE=tap100
+DISK_DEVICE=dummy.img
+
+INCLUDEOS_SRC=${INCLUDEOS_SRC-$HOME/IncludeOS}
+UNIKERNEL_SRC=${INCLUDEOS_SRC}/examples/demo_service
+UNIKERNEL_BUILD=${UNIKERNEL_SRC}/build_ukvm
+UNIKERNEL_IMG=${UNIKERNEL_BUILD}/IncludeOS_example
+ARCH=${ARCH:-x86_64}
+SOLO5_SRC=${INCLUDEOS_SRC}/build_${ARCH}/precompiled/src/solo5_repo
+
+die_error ()
+{
+    echo $0: "$@" 1>&2
+    exit 1
+}
+
+die_info ()
+{
+    echo $0: "$@" 1>&2
+    exit 0
+}
+
+SYSTEM=`uname -a`
+[[ ! $SYSTEM =~ .*[L|l]inux.* ]] && die_info "Solo5/ukvm is currently only supported on Linux."
+
+trap nuketmpdir 0 INT TERM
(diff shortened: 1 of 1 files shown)


ricarkol:
Actually, it seems that the jenkins job "Unit tests, examples and binary blobs tests" is trying to run the new test. It seems that it should be enough to rewrite the test in python (or at least python that starts this bash script).

fwsGonzo:
We'll put our top men on the case :)

ricarkol:
Thanks @fwsGonzo

alfred-bratterud:
Great @ricarkol - Jenkins should be able to run that just fine - it's used to starting VM's for running our integration tests so it has nested access to KVM / vt-x. @mnordsletten can you look into why the unit tests failed? In the meantime I'll try to trigger a second run.

alfred-bratterud:
jenkins test please

alfred-bratterud:
Oh, one thing - I'm not sure our testrunner can run this test as it's not using the python vmrunner. I do know that the testrunner will run `test.sh` for anything under `test/misc/<test name>/` so the quickest way to get immediate feedback on how Jenkins will manage running the test is to move this to e.g. `test/misc/ukvm/test.sh`.

mnordsletten:
Yeah, you are right @alfred-bratterud. The testrunner is very particular about what the folder structure should be.

Try the setup that is proposed with `test/misc/ukvm/test.sh` and the test should be run.

ricarkol:
Thanks @alfred-bratterud and @mnordsletten , getting another error while building:
```
20:06:03 * ./misc/ukvm                                                         [  FAIL  ]
20:06:03 * Process stdout: 
20:06:03 ERROR
20:06:03 /tmp/tmp.SYNYuIgENT/demo_service.log.0: 
20:06:03 > ~/workspace/workspace/pr_misc/IncludeOS/examples/demo_service/build_ukvm ~/workspace/workspace/pr_misc/test/misc/ukvm
20:06:03 > CMake Error: The source directory "/home/ubuntu/workspace/workspace/pr_misc/IncludeOS/examples/demo_service" does not appear to contain CMakeLists.txt.
```
Will take a look at `./misc/build_examples` and do the same here.

Issue #1590, opened 2017-11-25
clean the code on header.cpp file


mywtfmp3:
In [src/net/http/header.cpp](../blob/master/src/net/http/header.cpp) file, 
```c++
Header::Const_iterator Header::find(util::csview field) const noexcept {
  if (field.empty()) return fields_.cend();
  //-----------------------------------
  return
    std::find_if(fields_.cbegin(), fields_.cend(), [&field](const auto _) {
      return (_.first.length() == field.length())
        and std::equal(_.first.data(), _.first.data() + _.first.length(), field.data(), [](const auto a, const auto b) {
          return std::tolower(a) == std::tolower(b);
        });
    });
}
```

It's possible to get rid of  `_.fisrt.length == field.length` by using alternative `std::equal`.

`bool std::equal(InputIt1 first1, InputIt1 last1, InputIt2 first2, InputIt2 last2, BinaryPredicate p);`

And when comparing string and string_view, is it better to use `cbegin()`, `cend` consistently? Here is my code. 

```C++
Header::Const_iterator Header::find(util::csview field) const noexcept {
  if (field.empty()) return fields_.cend();
  //-----------------------------------
  return
    std::find_if(fields_.cbegin(), fields_.cend(), [&field](const auto _) {
      return std::equal(_.first.cbegin(), _.first.cend(), field.cbegin(), field.cend(), 
        [](const auto a, const auto b) { return std::tolower(a) == std::tolower(b); });
    });
}
```

AndreasAakesson:
You're right. Have you confirmed that your solution works? We do have a [unit test](https://github.com/hioa-cs/IncludeOS/blob/dev/test/net/unit/http_header_test.cpp) for the Header class (extend it if needed).

Feel free to make PR's where you see improvements :)

PR #1592, opened 2017-11-28
cleanup: use another std::equal function to remove verbose code

mywtfmp3:
#1590 Just a small change to fix the issue opened by myself.

However, I am not sure why `field` is capture by reference in the lambda function here,

`std::find_if(fields_.cbegin(), fields_.cend(), [&field](const auto _) {`

So I didn't make a change to this line.

The `field` is a const string_view, and is pass by value to `find` function.

BTW, can someone explain how unit test is organized in this repo? There is a `lest_util` folder. Am I supposed to use it?

### DIFF
--- a/src/net/http/header.cpp
+++ b/src/net/http/header.cpp
@@ -110,10 +110,8 @@ Header::Const_iterator Header::find(util::csview field) const noexcept {
   //-----------------------------------
   return
     std::find_if(fields_.cbegin(), fields_.cend(), [&field](const auto _) {
-      return (_.first.length() == field.length())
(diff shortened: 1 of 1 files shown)


ingve:
ok to test

ingve:
Lest (the unit test framework we use) should be ready-to-use after you've installed IncludeOS.

To build and run the unit tests:

```
$ cd test
$ mkdir build && cd build
$ cmake ..
$ make && ./unittests
```

ingve:
(after building the unittests binary, you can run `./unittests --pass` if you want output from the passing tests as well)

Issue #1613, opened 2017-12-28
IncludeOS/VirtualBox - vboxrun.sh should specify Linux 64-bit


mjbright:
As per #1611 I've tried to follow the wiki instructions to run an IncludeOS unikernel on VirtualBox.

Using vboxrun.sh rather than the VirtualBox GUI also fails horribly with
```emR3Debug: rc=VERR_VMX_INVALID_GUEST_STATE```

because IncludeOS now buids 64-bit not 32-bit.

vboxrun.sh should specify "Other_64" as os type at line 82:
```    $VB createvm --name "$VMNAME" --ostype "Other_64" --register ```

PR #1614, opened 2017-12-28
Update vboxrun.sh to specify "Other_64" as ostype

mjbright:
IncludeOs now builds 64-bit images, so ostype needs updating (#1613)


AndreasAakesson:
ok to test

PR #1615, opened 2017-12-30
Fix GCC 5.4 compiler error (to master)

forderud:
Fixes the following error reported in https://github.com/hioa-cs/IncludeOS/issues/1605:
```
IncludeOS/examples/snake/snake.hpp:179:27: error: cannot call member function ‘void Snake::game_loop()’ without object
   [this](auto) { game_loop(); }
```

I'm already aware that I'm not supposed to build IncludeOS with GCC. Still, I think it's beneficial to limit dependence to a specific compiler.

### DIFF
--- a/examples/snake/snake.hpp
+++ b/examples/snake/snake.hpp
@@ -176,7 +176,7 @@ void Snake::game_loop()
 
 	Timers::oneshot(
 		std::chrono::milliseconds(_head_dir.x() == 0 ? 120 : 70),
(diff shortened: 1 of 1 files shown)


ingve:
ok to test

forderud:
Closing, since the PR is accidentally targeted towards "master" instead of "dev". Please see https://github.com/hioa-cs/IncludeOS/pull/1616 instead.

Issue #1607, opened 2017-12-16
Configure network card by MAC instead of index


perbu:
Configuring the network card by "index" is error-prone. Also it is not obvious what card is what. How about the option to tie an interface definition to a card on the basis of the MAC address.

This makes it a lot easier to identify the networking card in the hypervisor and thereby making sure it is connected to the right networks.

The syntax would be simple. Just replace "index" with "mac" and the address.

PR #1619, opened 2018-01-04
Get/create stack by MAC addr in Super_stack

AndreasAakesson:
To support #1607. Uplink now respects index (can be either number or MAC).


AndreasAakesson:
jenkins test please

PR #1624, opened 2018-01-06
Removed '../run.sh' from Readmes.

xekoukou:
I think this is ok. I couldn't start the acorn example with ```../run.sh```.


ingve:
Hi @xekoukou, thanks for contributing! Could you redirect this PR to the `dev` branch?

PR #1632, opened 2018-01-12
Fix softreset location

fwsGonzo:
- Would cause the OS to think it was not live updated, when in fact it was


fwsGonzo:
jenkins test please

fwsGonzo:
jenkins test please

PR #1644, opened 2018-01-17
Emergency VGA logbuffer driver 

fwsGonzo:
screenshot from 2018-01-17 22-20-28
New driver: vga_emergency
- Does not need the vga output driver to work and uses no heap
- Logs to 128k ringbuffer
- Re-activates interrupts on panic and enables keyboard handler
- Prints partial VGA buffer based on position in buffer
- Navigate up and down the OS log while in panic!

### DIFF
--- a/api/hw/ps2.hpp
+++ b/api/hw/ps2.hpp
@@ -58,8 +58,6 @@ namespace hw
       VK_COUNT
     };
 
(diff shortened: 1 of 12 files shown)


fwsGonzo:
jenkins test please

PR #1661, opened 2018-02-03
ip4: Refactored code in cidr.hpp

RicoAntonioFelix:


RicoAntonioFelix:
@AndreasAakesson Since you've initially marked that the object can be constructed within a compile time context, it made sense to also have the getters be used in such a context also.

PR #1693, opened 2018-02-21
HTTP client_connection.cpp: Replacing Ensures(on_response_) with if statement

AnnikaH:
To avoid panic if the request has timed out and the response is received after this


AndreasAakesson:
Looks like everything was great with the PR, but it just timed out.
```
14:08:50 Waiting for the completion of pr_stress
14:22:07 Build timed out (after 20 minutes). Marking the build as aborted.
14:22:07 Build was aborted
```

Looking at [pr_stress](https://jenkins.includeos.org/job/pr_stress/) I can't see that any of the jobs has failed.

mnordsletten:
Jenkins test please

PR #1695, opened 2018-02-23
etc: ensure (already installed) clang is at least 5.0

justinc1:
With "export CC=/usr/bin/clang-3.8" and actually installed clang-3.8,
we failed to request at least 5.0 version.

The "export CC=" line was added as install.sh suggested, so I
guess it is quite common case.

Signed-off-by: Justin Cinkelj <justin.cinkelj@xlab.si>


ingve:
ok to test

fwsGonzo:
jenkins test please

mnordsletten:
Let's give this another go
Jenkins test please

PR #1696, opened 2018-02-25
Fix leading tab in linux install script

fredrb:
When listing the dependencies for Linux installation - there is leading tab on the *installed* items. As seen here:
include_os_2

This pull requests proposes the following:
- Remove all leading tabs from both *OS* dependencies and *pip* dependencies.
- Adds the header *Status  Package  Version* to *pip* dependencies
- Refactor *installed/missing* text coloring with helper functions

Result: 
include_os_result


ingve:
Thanks for contributing!

ok to test

fredrb:
It seems like the building is failing in Jenkins but its being installed successfully and tests are passing.

```
14:29:49 pr_stress #467 completed. Result was SUCCESS
```

mnordsletten:
Yeah, this test failed due to an internal bug. It should be resolved now.  I'll trigger the test to start again. 

Jenkins test please

PR #1697, opened 2018-02-25
"system_log" logging interface + plugin implementation

fwsGonzo:
* `stdout` behavior is by default directly to OS::print, as expected
* When system_log plugin enabled stdout is routed to the system log
* System log by default only prints during panic()
* System log interface in `api/system_log`
* RB is preserved during LiveUpdate, zero overhead
* LiveUpdate test median time: 30ms before, 27.81ms after

### DIFF
--- a/api/fs/memdisk.hpp
+++ b/api/fs/memdisk.hpp
@@ -28,6 +28,11 @@ namespace fs {
   public:
     static constexpr size_t SECTOR_SIZE = 512;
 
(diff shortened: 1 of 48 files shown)


fwsGonzo:
while (not green) jenkins test please

fwsGonzo:
jenkins test please

fwsGonzo:
jenkins test please

fwsGonzo:
jenkins test please

PR #1703, opened 2018-03-01
uplink: No longer delay liveupdate, but instead flush

AndreasAakesson:
Bandaid for when timers are broken.


AndreasAakesson:
jenkins test please

PR #1706, opened 2018-03-05
Allow booting of solo5/ukvm with the boot command

niks3089:
Changes done:

- The IncludeOS can now be booted on solo5 as follows
`boot --with-solo5 .`
- Changed test_ukvm.sh to reflect this change.
- vm.json file is not required for solo5 currently but can be used if required in the future.
-  Created a new class ukvm under hypervisor
- Moved start_process() function to the parent class(hypervisor). Similarly, stop(), wait() and other functions could be moved to the parent class.

---
This change is [Reviewable](https://reviewable.io/reviews/hioa-cs/includeos/1706)

### DIFF
--- a/vmrunner/vmrunner.py
+++ b/vmrunner/vmrunner.py
@@ -179,6 +179,136 @@ class hypervisor(object):
     def image_name(self):
         abstract()
 
+    def start_process(self, cmdlist):
+
+        if cmdlist[0] == "sudo": # and have_sudo():
+            print color.WARNING("Running with sudo")
+            self._sudo = True
+
+        # Start a subprocess
+        self._proc = subprocess.Popen(cmdlist,
+                                      stdout = subprocess.PIPE,
+                                      stderr = subprocess.PIPE,
+                                      stdin = subprocess.PIPE)
+
+        return self._proc
+
+
+# Ukvm Hypervisor interface
+class ukvm(hypervisor):
+
+    def __init__(self, config):
+        # config is not yet used for ukvm
+        super(ukvm, self).__init__(config)
+        self._proc = None
+        self._stopped = False
+        self._sudo = False
+        self._image_name = self._config if "image" in self._config else self.name() + " vm"
+
+        # Pretty printing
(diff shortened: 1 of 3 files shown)


AndreasAakesson:
Cool! 
ok to test

vmrunner/vmrunner.py
```diff
@@ -180,6 +180,136 @@ def name(self):
...
+        self.info = Logger(color.INFO("<" + type(self).__name__ + ">"))
+
+    def name(self):
+        return "Ukvm"
+
+    def image_name(self):
+        return self._image_name
+
+    def drive_arg(self):
+        return ["--disk=" + INCLUDEOS_HOME + "dummy.disk"]
```
ricarkol: Some services need a real disk image, like https://github.com/hioa-cs/IncludeOS/tree/master/test/fs/integration/fat32.

niks3089: I see, I guess we can use vm.json to pass the disk and net parameters when not present defaults to dummy.disk and tap100. 
Also, instead of setting tap interface and creating dummy.disk in boot.py, we can move that to vmrunner.py where it can be configured based on vm.json

A sample vm.json
```
{
  "net" : "tap101"
  "disk"   : "another.disk"
}
``` 
Thoughts?

fwsGonzo:
Thank you for contributing to IncludeOS! This is a much needed feature for our solo5 buffs.

It's hard to follow your commit messages, have you read our guidelines?
http://includeos.readthedocs.io/en/latest/Contributing-to-IncludeOS.html

The easiest fix would be to rebase everything into a single commit, something ala "boot: Add --with-solo5 option". Also, you can add the disk support in a separate commit later.

vmrunner/vmrunner.py
ricarkol: Was thinking what to do about the TAPs. Qemu runs get a script that's used to create the TAP. In the case of solo5/ukvm, it seems that vmrunner should be the place where the TAPs are created (and the disk).

niks3089:
@fwsGonzo : Re-based all the changes as a single commit for this PR as well for PR #1707

PR #1707, opened 2018-03-05
Udp performance test

niks3089:
Changes done:

- Runs a udp server or a client as a IncludeOS unikernel
- When run as a server, sample is calculated every 5 seconds. 
- When run as a client, a bunch of packets is sent every second for default 10 seconds. Need to pass "client" argument to the program to run it as a client.
- Uses ncat to send and receive packets. 
- Works with Qemu for now. With ukvm, there is a timing issue resulting in samples not being collected correctly

---
This change is [Reviewable](https://reviewable.io/reviews/hioa-cs/includeos/1707)

### DIFF
--- a/examples/UDP_perf/CMakeLists.txt
+++ b/examples/UDP_perf/CMakeLists.txt
@@ -1,0 +1,35 @@
+cmake_minimum_required(VERSION 2.8.9)
+
+# IncludeOS install location
+if (NOT DEFINED ENV{INCLUDEOS_PREFIX})
+  set(ENV{INCLUDEOS_PREFIX} /usr/local)
+endif()
+include($ENV{INCLUDEOS_PREFIX}/includeos/pre.service.cmake)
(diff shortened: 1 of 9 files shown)


ingve:
ok to test

ingve:
Some tests seem to be timing out, not sure what is happening there. We'll have to invite @mnordsletten over to investigate.

mnordsletten:
Yeah, I've verified that the tests do actually pass, just that they are timing out easily. We'll give them another go, and I've eased up on some of our timeouts. 

Jenkins test please

niks3089:
@fwsGonzo , I've removed the NaCl module, however, I am unable to squash and re-base all the commits as a single commit.  Some of the commits were not reflected when I tried this. Hopefully this should be ok.

niks3089:
@AndreasAakesson, this is done now. Please take a look

Issue #1721, opened 2018-03-20
possible deca-millennium bug in isotime::to_datetime_string


ingve:
If I run the following code:

```
time_t add_hour(time_t t) {
	return t + (60*60);
}

int main(int argc, char *argv[]) {
	time_t t0 = 0L;
	time_t t1 = add_hour(t0);
	time_t t2 = 253402297200L;
	time_t t3 = add_hour(t2);
	printf("%s\n", isotime::to_datetime_string(t0).c_str());
	printf("%s\n", isotime::to_datetime_string(t1).c_str());
	printf("%s\n", isotime::to_datetime_string(t2).c_str());
	printf("%s\n", isotime::to_datetime_string(t3).c_str());	
}
```

in IncludeOS (v0.12.0-rc.3-271-g25f83a3d-dirty), I get the following output:

```
1970-01-01T00:00:00Z
1970-01-01T01:00:00Z
9999-12-31T23:00:00Z
+10000-01-01T00:00:0
```

This seems to be caused by not checking the return value from `strftime` inside `isotime::to_datetime_string`

> The number of bytes written into the character array pointed to by str not including the terminating '\0' on success. If count was reached before the entire string could be stored, ​0​ is returned and the contents are undefined.

— [strftime@cppreference.com](http://www.cplusplus.com/reference/ctime/strftime/)

0 is indeed returned in the last case above. Checking the return value would be easy enough, but I don't know what would be a reasonable string to return from `to_datetime_string` in this case (truncated string? empty string? "invalid date" error message? throwing an exception?) [¯\\\_(ツ)_/¯](https://rachelbythebay.com/w/2013/03/17/time/).

AndreasAakesson:
It now returns a empty string on error. Good enough?

PR #1723, opened 2018-03-22
Fixed unittest on mac + isotime issue

AndreasAakesson:
Building unittest on mac now uses brew bundled libc++. The user still has to export correct CC/CXX for it to build (Apple do not support c++17 on native Sierra 10.12.6).

isotime now checks the return value from strftime and return empty string if something is wrong. #1721

Bufferstore now using aligned_alloc instead of memalign (supposedly better and less mocking for mac unittest). Also replaced `next_` with unique_ptr due to some weird missing stuff on mac. The problem didnt go away but the error was cleaner. 
What fixed the problem was using brew bundled libc++.

### DIFF
--- a/api/net/buffer_store.hpp
+++ b/api/net/buffer_store.hpp
@@ -93,7 +93,7 @@ namespace net
     size_t               bufsize_;
     uint8_t*             pool_;
     std::vector<uint8_t*> available_;
-    BufferStore*         next_;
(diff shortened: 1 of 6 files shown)


AndreasAakesson:
jenkins test please

AndreasAakesson:
jenkins test please

AndreasAakesson:
jenkins test please..

PR #1708, opened 2018-03-06
New interrupt stacks etc.

alfred-bratterud:
Ref #1704

PR #1724, opened 2018-03-23
Fix booting of solo5 using '--with-solo5' option

niks3089:
When I tried to run IncludeOS with solo5 with `boot --with-solo5 .`  I got the following error

```
Traceback (most recent call last):
  File "/home/osboxes/includeos//includeos/vmrunner/vmrunner.py", line 855, in boot
    self._hyper.boot(multiboot, debug, "booted with vmrunner", "udp_perf")
TypeError: boot() takes at most 4 arguments (5 given)
```

Looks like this because of a recent change (#1708) where the `debug` parameter was added to Qemu but not to ukvm.  I simply added the debug parameter and its a NO-OP for now

---
This change is [Reviewable](https://reviewable.io/reviews/hioa-cs/includeos/1724)


niks3089:
Oops, PR shows multiple commits even though only a single commit with a single file has changed.

PR #1728, opened 2018-03-27
HTTPS supported client

AndreasAakesson:
Also replaced URI storage from `std::string` to `vector<char>`. Using a string made it unreliable when moving, due to string being copied and not moved.


AndreasAakesson:
jenkins test please 🙏

AndreasAakesson:
jenkins test please

PR #1759, opened 2018-05-10
ip6: Initial icmpv6 support and very basic NDP support along with tests

niks3089:
Summarising the changes done: 

Inet: Added icmp6 handler, new network_config6 API
IP6:  Address bug fixes, changing prefix as an int instead of v6 address, new unit tests
ICMP6: New icmp6 APIs and support for icmp6 types, codes & errors and sending and receiving icmp packets
NDP: Neighbour solicitation and advertisement support. 
IPv6 bridge: Bridge will configured with v6 gateway during creation
Tests: Added few more unit tests and icmp6 integration test

---
This change is [Reviewable](https://reviewable.io/reviews/hioa-cs/includeos/1759)


niks3089:
@AndreasAakesson :  I tried this when I was initially re-factoring icmp6. I moved the icmp4_common & icmp6_common to a single file under the same namespace.  Note that we still need to have separate type & code class for each version due their differences. v6 has different types with different code  so we would still need to have 2 get_type_string and get_code_string functions so icmp*_common file will essentially be merging 2 files into 1 without any re-usability.  With icmp_error we could use templates for v4/6 type and code but I don't think we gain much from this.  I thought it would be better to maintain and have 2 separate files in each module due to more differences than similarities.

niks3089:
Couple of commits to be able to send and receive icmp6 pings to & from IncludeOS.

alfred-bratterud:
Nice job @niks3089, with this I'm successfully getting ping6 replies from IncludeOS and icmp6 replies from the host to IncludeOS. Please add a `test.py` to enable jenkins to run the icmp6 integration test. For now though I've confirmed that it works locally and merged.

PR #1760, opened 2018-05-11
Changing IP addresses for dhcpd integration test

staiyeba:
Changing IP addresses due to failed integration tests on new test workers.
The earlier network mask was more open so a larger IP range worked.
After recent change to the network mask, the dhcpd integration tests failed.
Thus changing IP address to a smaller range for the network tests.


ingve:
add to whitelist

mnordsletten:
Jenkins test please

PR #1767, opened 2018-05-16
 Add test.py to icmp6 integration test. 

niks3089:
This change is [Reviewable](https://reviewable.io/reviews/hioa-cs/includeos/1767)

### DIFF
--- a/test/net/integration/icmp6/test.py
+++ b/test/net/integration/icmp6/test.py
@@ -1,0 +1,50 @@
+#! /usr/bin/env python
+
+import sys
+import os
+import subprocess
+
+includeos_src = os.environ.get('INCLUDEOS_SRC',
+                               os.path.realpath(os.path.join(os.getcwd(), os.path.dirname(__file__))).split('/test')[0])
+sys.path.insert(0,includeos_src)
+
+from vmrunner import vmrunner
+from vmrunner.prettify import color
+
+import socket
+
+# Get an auto-created VM from the vmrunner
+vm = vmrunner.vms[0]
+
+num_successes = 0
+
+def start_icmp_test(trigger_line):
+  global num_successes
+
+  # 1 Ping: Checking output from callback in service.cpp
+  print color.INFO("<Test.py>"), "Performing ping6 test"
+
+  output_data = ""
+  for x in range(0, 11):
+    output_data += vm.readline()
+
+  print output_data
+
+  if "Received packet from gateway" in output_data and \
+    "Identifier: 0" in output_data and \
+    "Sequence number: 0" in output_data and \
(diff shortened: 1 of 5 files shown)


mnordsletten:
The `Intrusive` key in the `vm.json` indicates that the test performs some sort of package/environment installation. I don't think this is necessary here.

Also in my pr: https://github.com/hioa-cs/IncludeOS/pull/1768 I've tried to resolve a few ip address conflicts that might arise when running tests in parallel. I know that `10.0.0.45` is already in use by the normal icmp test. I know that at the moment `10.0.0.51` is not in use by any other test, so you could change to use that instead.

niks3089:
Thanks @mnordsletten, I have made the changes. Also, I was looking at the network tests console ouptut and I see that the tests marked as intrusive are skipped. So they are never run on jenkins?

mnordsletten:
Great :) 

As it is right now they are run as a separate job every night on a newly spawned machine. https://jenkins.includeos.org/job/pr_intrusive_tests

mnordsletten:
I might have made a mistake regarding `10.0.0.51` It seems like the vlan tests are actually using that address. Sorry about that. I know for sure that `10.0.0.52` is available though.

test/net/integration/icmp6/test.py
```diff
@@ -0,0 +1,50 @@
...
+    "Type: ECHO REPLY (129)" in output_data and \
+    "Code: DEFAULT (0)" in output_data and \
+    "Checksum: " in output_data and \
+    "Data: INCLUDEOS12345ABCDEFGHIJKLMNOPQRSTUVWXYZ12345678" in output_data:
+    num_successes += 1
+    print color.INFO("<Test.py>"), "Ping test succeeded"
+  else:
+    print color.FAIL("<Test.py>"), "Ping test FAILED"
+
+vm.on_output("Service IPv4 address: 10.0.0.45, IPv6 address: fe80:0:0:0:e823:fcff:fef4:85bd", start_icmp_test);
```
mnordsletten: Make sure you change this line here to match the `10.0.0.52` address as well.

mnordsletten:
I did some checking and ipv6 is indeed working. I got your test to work, but a few modifications are needed:

1. There aren't 11 lines to read output from. We got it working with range(0, 9). This might need to be more dynamic, if the number of output lines changes in the future?
2. Needs to call vm.Exit(0, "Test completed without errors"). The vm does not shut down automatically. This call can be done after print "ping test succeeded"

I think a little rewrite of the `start_icmp_test` logic might be necessary :)

niks3089:
I need to add more tests. I've simply made the required fixes to get the tests to pass. I will try to make it dynamic when I add more tests

PR #1768, opened 2018-05-16
Tests: Resolved potential ip conflicts in all network tests

mnordsletten:
- Moved overview of which ip's are in use to `test/README.md` so that it can be edited by any test that needs to use an ip address. 
- Silenced output from a lot of setup commands.


AndreasAakesson:
Nice! There are also some networks test in `test/posix/`.

PR #1770, opened 2018-05-21
CONTRIBUTE.md: Fix link to contribution guidelines

knatten:
The old link was to a wiki that seems to no longer exist. In README.md,
I found a link to the contribution guidelines on readthedocs, so I
assume that's the one we want to use.


ingve:
add to whitelist

PR #1778, opened 2018-05-25
Test: Suppress output from apt command in router

mnordsletten:
Also created to test new jenkins build machine.


mnordsletten:
Jenkins test please

PR #1781, opened 2018-05-28
Fix compile error when using test_ukvm.sh

mhk:
Moved `extern "C"` to correct location. Otherwise, when executing `./test_ukvm.sh` the following compile error is generated:
```
includeos/src/service_name.cpp:24:8: error: expected unqualified-id before string constant
 extern "C"
        ^
```

### DIFF
--- a/src/service_name.cpp
+++ b/src/service_name.cpp
@@ -20,8 +20,8 @@
 extern "C" __attribute__((noreturn))
 void panic(const char* reason);
 
-__attribute__((noreturn))
-extern "C" void abort(){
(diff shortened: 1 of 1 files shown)


ingve:
ok to test

mhk:
Can someone tell me if this is a sporadic?

ingve:
Hi @mhk, our Jenkins guru par excellence, @mnordsletten, is investigating.

mnordsletten:
We do indeed have a pretty sporadic failure rate at the moment. Improvements are in the works. 

I got this PR to pass in the end 👍

PR #1782, opened 2018-05-28
Various

fwsGonzo:


mnordsletten:
Please use `apt-get -q` in the setup.sh to mute any output

PR #1792, opened 2018-06-01
net: Clear headers in a more standard way 

fwsGonzo:
This PR contains a change that replaces memset with copy assignment when initializing Packet headers. It's a more standard approach.

Also, build linux platform with GCC and add README.


fwsGonzo:
Of course, I copied verbatim from IPv4 which uses older vernacular :)

PR #1801, opened 2018-06-07
Unittest for FAT

fwsGonzo:
and some WIP code for unittesting filesystem


fwsGonzo:
Not sure how to "fix" this other than copying the FAT file into the build folders just so it can find the file..

PR #1808, opened 2018-06-14
cmake: Dont strip image unless expl enabled

fwsGonzo:


fwsGonzo:
It seems this change provoked a bug that I cant replicate locally. Its happening in icmp6, with the integration test.

PR #1818, opened 2018-06-24
Buddy allocator

alfred-bratterud:
Replace the stack allocator with buddy allocator as mmap backend


alfred-bratterud:
Jenkins test please

PR #1820, opened 2018-06-26
uplink: use UUID as id

ingve:


fwsGonzo:
UUID is already supported in vm.json

PR #1825, opened 2018-06-27
Delegate changes

### LINKED ISSUE #1821: Moving a delegate does not make it "empty"
```
delegate<void()> del;
// fine
assert(not del);
assert(del == nullptr);

del = [](){};
// fine
assert(del);
assert(del != nullptr);

auto del2 = std::move(del);
// fine
assert(del2);
assert(del2 != nullptr);
// not fine
assert(not del);
assert(del == nullptr);
```

```
class delegate {
...
	bool operator== (std::nullptr_t) const noexcept
	{
		return storage_.empty();
	}

	bool operator!= (std::nullptr_t) const noexcept
	{
		return !storage_.empty();
	}

	explicit operator bool() const noexcept
	{
		return !storage_.empty();
	}
...
};
```
I'm pretty sure this has been the case (since a lot of logic have been built around it). This isn't verified in the unittest tho (needs to be!).
There are no custom move ctor/assignment what I can see (= default), but moving the storage seems to not making the other storage empty.

AndreasAakesson:
Checking if this breaks anything. Not sure if it has any side effects to make the destructor_ptr to nullptr, since the destructor of the inplace delegate checks if it exists before calling it. This will kinda fix #1821, but this is only for the inplace type. Could be that it's necessary to create custom move construct & assignment for the other types as well.

### DIFF
--- a/api/net/tcp/stream.hpp
+++ b/api/net/tcp/stream.hpp
@@ -101,7 +101,6 @@ namespace net::tcp
     void close() override
     {
       auto onclose = std::move(this->m_on_close);
-      this->m_on_close.reset();
       m_tcp->reset_callbacks();
       m_tcp->close();
       if (onclose) onclose();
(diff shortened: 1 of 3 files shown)


fwsGonzo:
Lets try this and see what happens...

PR #1828, opened 2018-06-27
Additional g++ support

fwsGonzo:
I dont know of any services that dont run right now, so while the support is brittle, its good enough to run all the tests and examples. Perhaps except STREAM, which is very slow. But maybe its supposed to be slow.

*hashtag road to g++*


fwsGonzo:
Any further issues now should be easier to solve, as most if not all services are up and running.

PR #1834, opened 2018-06-29
Various

fwsGonzo:
This PR can be rerun until it fails. There are examples which shouldnt build right now.


fwsGonzo:
This PR apparently fixes issues when booting on GCE, mostly due to misaligned LU virtual storage area

PR #1841, opened 2018-07-05
Let DHCP continue forever if no timeout

### LINKED ISSUE #1716: Giving up early on DHCP
The OS now gives up on DHCP and never tries again even though it may just be a temporary problem. Secondly, the OS may be statically configured after a certain time passes. This is mostly because we run qemu with bridge, even though there are other network options that provide DHCP.

We need to explore:
- Retrying DHCP forever
- Rebooting after repeated failed contact on uplink, as no one wants a "bricked" instance
- Changing qemu bridged mode to use eg. NAT, host-local or other DHCP-enabled modes

AndreasAakesson:
By supplying timeout = 0, DHCP will now continue forever.
I added to constants that can be tuned:
```
    static constexpr std::chrono::seconds RETRY_FREQUENCY{1};
    static constexpr std::chrono::seconds RETRY_FREQUENCY_SLOW{10};
```
There is also hardcoded that it should make 10 tries before slow face:
```
  void DHClient::restart_negotation()
  {
    ...
    static const int FAST_TRIES = 10;
    // never timeout
    if(tries <= FAST_TRIES)
    {
      // do fast retry
      timeout_timer_.start(RETRY_FREQUENCY);
    }
    else
```
Fixes #1716. Inet still got default timeout to 10 seconds (also not using std::chrono), same if nothing is supplied in config.json. For it to work, config.json needs the field `timeout: 0` to try forever.
So i guess that's todo/decision to be made what the default should be.

### DIFF
--- a/api/net/dhcp/dh4client.hpp
+++ b/api/net/dhcp/dh4client.hpp
@@ -30,7 +30,8 @@ namespace net {
   class DHClient
   {
   public:
-    static const int NUM_RETRIES = 5;
+    static constexpr std::chrono::seconds RETRY_FREQUENCY{1};
+    static constexpr std::chrono::seconds RETRY_FREQUENCY_SLOW{10};
 
     using Stack = IP4::Stack;
     using config_func = delegate<void(bool)>;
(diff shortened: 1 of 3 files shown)


AndreasAakesson:
jenkins test please

PR #1846, opened 2018-07-20
catch exception by const reference

### LINKED ISSUE #1845: Catch exception by const reference
Some code catches exception by value or by non-const reference. Change them to const reference.

mywtfmp3:
Fix #1845


ingve:
ok to test

PR #1850, opened 2018-07-30
solo5: Make event loop tickless

fwsGonzo:
Event loop now waits for next timer, when one is present, otherwise tick 500ms (change in solo5/os.cpp). Since all normal services use timers, the end result is a tickless kernel. Previously the kernel ticked every 0.5 milliseconds, which was overmuch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ set(CAPABS "${CAPABS} -fstack-protector-strong -D_STACK_GUARD_VALUE_=0x${STACK_P
 # Various global defines
 # * NO_DEBUG disables output from the debug macro
 # * OS_TERMINATE_ON_CONTRACT_VIOLATION provides classic assert-like output from Expects / Ensures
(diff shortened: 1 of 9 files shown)


fwsGonzo:
@ricarkol Does the assumptions in the new tickless event loop seem right?

ricarkol:
> Does the assumptions in the new tickless event loop seem right? This is correct:
`res = solo5_yield(solo5_clock_monotonic() + nxt.count());`
If `nxt.count()` is 0, the call will eventually be `sys_poll(timeout=0)` which does what you expect.

>  I could have waited forever when no timers were present
You could wait forever when no timers are present, and the system will just do that: wait forever (until some packet arrives). I think what you have (sleep for some time) is the right behavior. Think of this program: a TCP sender that never gets an ACK because the receiver died. The issue is that this sender will never time out because it's sleeping forever waiting for a packet (that will never arrive).

ricarkol:
Actually, in the example above. This sender _will_ have a timer. Can you think of a program that leads to a situation like that: solo5_yield is called without any timer.

fwsGonzo:
Alright, so I tried using 0 for `solo5_yield` and it did not wait forever, so we are back to just using longer waits (which is fine). I also noticed that events were not being processed, so I fixed that. And finally, I modified TCP_perf to work on solo5 platform as well.

PR #1851, opened 2018-07-31
LiveUpdate example is now an IRC server

fwsGonzo:
.. that can be live updated


alfred-bratterud:
I want to be on that IRC server pls :-)

PR #1852, opened 2018-08-01
Ndp and Mld changes

niks3089:
MLD changes are very basic and still needs a lot more work

---
This change is [Reviewable](https://reviewable.io/reviews/hioa-cs/includeos/1852)

### DIFF
--- a/api/net/inet.hpp
+++ b/api/net/inet.hpp
@@ -81,20 +81,20 @@ namespace net {
     hw::Nic& nic() const
     { return nic_; }
 
-    IP4::addr ip_addr() const
-    { return ip4_addr_; }
+    IP4::addr ip_addr()
+    { return ip_obj().ip4_addr(); }
 
-    IP4::addr netmask() const
-    { return netmask_; }
+    IP4::addr netmask()
+    { return ip_obj().ip4_netmask(); }
 
-    IP4::addr gateway() const
-    { return gateway_; }
+    IP4::addr gateway()
+    { return ip_obj().ip4_gateway(); }
 
     IP4::addr dns_addr() const
     { return dns_server_; }
 
-    IP4::addr broadcast_addr() const
-    { return ip4_addr_ | ( ~ netmask_); }
+    IP4::addr broadcast_addr()
+    { return ip_obj().broadcast_addr(); }
 
     const ip6::Addr& ip6_addr() const noexcept
     { return ip6_addr_; }
@@ -257,7 +257,7 @@ namespace net {
 
     void set_gateway(IP4::addr gateway)
     {
-      this->gateway_ = gateway;
+      this->ip_obj().set_gateway(gateway);
     }
 
     void set_dns_server(IP4::addr server)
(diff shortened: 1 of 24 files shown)


api/net/inet.hpp
```diff
@@ -277,16 +281,16 @@ namespace net {
 
     /* Automatic configuration of ipv6 address for inet */
     void autoconf_v6(int retries = 0, slaac_timeout_func = nullptr,
-            IP6::addr alternate_addr = IP6::ADDR_ANY);
+            ip6::Addr alternate_addr = IP6::ADDR_ANY);
 
-    bool is_configured() const
```
AndreasAakesson: Many of these functions can still be `const` if you refer to the member instead of the non const `ip_obj()` func (same goes for `ndp()` I guess?).

api/net/ip4/ip4.hpp
```diff
@@ -306,7 +306,32 @@ namespace net {
     PMTU minimum_MTU() const noexcept
     { return (PMTU) PMTU_plateau::ONE; }
 
+    ip4::Addr ip4_addr() const noexcept
```
AndreasAakesson: Uncertain of these functions need to contain the `ip4_` prefix due to being a member in IP4. What do you think?

src/net/ip6/mld.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+  Mld::MulticastHostNode::MulticastHostNode()
+  {
+    state_handlers_[Mld::HostStates::NON_LISTENER] =
+        [this] (icmp6::Packet& pckt)
+        {
+        };
+
+    state_handlers_[Mld::HostStates::DELAYING_LISTENER] =
+        [this] (icmp6::Packet& pckt)
```
AndreasAakesson: These delegates could be member functions :)

src/net/ip6/mld.cpp
```diff
@@ -0,0 +1,241 @@
...
+#include <vector>
+#include <net/inet>
+#include <net/ip6/mld.hpp>
+#include <net/ip6/packet_mld.hpp>
+#include <net/ip6/icmp6.hpp>
+#include <statman>
+
+namespace net
+{
+  Mld::Mld(Stack& inet) noexcept:
```
AndreasAakesson: Minor stuff, but would you mind typing these constructors differently?
```
Mld::Mld(Stack& inet) noexcept
  : inet_{inet},
  host_{*this},
  router_{*this}
```
Is the most used one, but I guess everything works as long as it's indentation on the member initialization.

src/net/ip6/ndp.cpp
```diff
@@ -427,6 +530,45 @@ namespace net
...
+  void Ndp::check_neighbour_reachability()
+  {
+  }
+
+  void Ndp::delete_dest_entry(ip6::Addr ip)
+  {
+    //TODO: Better to have a list of destination
+    // list entries inside router entries for faster cleanup
+    std::vector<IP6::addr> expired;
+    for (auto ent : dest_cache_) {
```
AndreasAakesson: The pattern you can use here is:
```
for(auto it = dest_cache_.begin(); it != dest_cache_.end(); )
{
  if(it->second.next_hop() == ip)
    it = dest_cache_.erase(it);
  else
    it++;
}
```

PR #1861, opened 2018-08-08
test: extending timeout to 40s for kernel/paging

staiyeba:


alfred-bratterud:
@staiyeba this shouldn't be necessary after all - the last time it timed out it was due to an actual error :-)

PR #1862, opened 2018-08-08
refactor emplace back

mywtfmp3:
It's not exception safe to do this,
```cpp
vector<unique_ptr<T>> v;
v.emplace_back(new T());
```

as `emplace_back` could cause a leak of this pointer if it throws a exception before emplacement (e.g. not enough memory to add a new element)


ingve:
add to whitelist

PR #1867, opened 2018-08-14
fix new without delete

mywtfmp3:
Called `new` in constructor
```cpp
new epoll_event;
```
but didn't release the resource in destructor.

### DIFF
--- a/linux/src/drivers/tap_driver.cpp
+++ b/linux/src/drivers/tap_driver.cpp
@@ -136,6 +136,7 @@ TAP_driver::TAP_driver(const char* devname,
 
 TAP_driver::~TAP_driver()
 {
+  delete epoll_ptr;
   close (this->tun_fd);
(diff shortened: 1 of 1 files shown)


fwsGonzo:
Perhaps it would have been better off as a unique_ptr, but this will do for now :+1:

mywtfmp3:
Or can we just use `epoll_event` directly without pointer? There's only one event in tap driver for file descriptor `tun_fd`.

I was checking how IncludeOS uses `epoll` and surprised to find it allocate the `epoll_event` on the heap for one event. That's why I found the bug.

fwsGonzo:
Yea, you are just looking at my pet project - running the OS on linux. I dont have much experience with epoll, but I wanted to use it to make sure the performance was at least decent/comparable to running in a VM. I also have trouble using more than one TAP device. PLEASE HELP :pray:

PR #1884, opened 2018-08-31
clang-tidy performance-faster-string-find

mywtfmp3:
> Optimize calls to std::string::find() and friends when the needle passed is a single character string literal. The character literal overload is more efficient.

Runs clang-tidy over the compilation database. 

I may push more changes from clang-tidy results.

### DIFF
--- a/src/kernel/terminal.cpp
+++ b/src/kernel/terminal.cpp
@@ -132,7 +132,7 @@ split(const std::string& text, std::string& command)
   if (text.empty()) return retv;
   // extract command
   {
-    x = text.find(" ");
+    x = text.find(' ');
     // early return for cmd-only msg
     if (x == std::string::npos)
     {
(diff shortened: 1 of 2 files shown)


AndreasAakesson:
Nice :) Looking at `clang-tidy` now. Was it easy to setup and run on the whole project? Do you have an idea of how we could make this kinda automated with a script or something?

fwsGonzo:
I already wrote a script for this some time ago, and I believe I put it in the root dir, but it's not there anymore. I must have deleted it... You can look at the linux cmake scripts to see how to build the OS with tidy.

AndreasAakesson:
Yeah it figured it out with the help of the docs, already had the tooling. On Mac:
```
ccmake ..
# turn on CMAKE_EXPORT_COMPILE_COMMANDS
make
/usr/local/opt/llvm@6/share/clang/run-clang-tidy.py <options> # homebrew llvm
```

PR #1903, opened 2018-09-21
Build Docker Image: from IncludeOS repo

staiyeba:
Build Docker Image from within IncludeOS repository.


alfred-bratterud:
Thanks @staiyeba - @mnordsletten will review. The content of the docker readme should probably go under the documentation :-)

.dockerignore
```diff
@@ -0,0 +1,9 @@
+build_*
+IncludeOS_install/
+diskimagebuild/build
+vmbuild/build
+src/chainload/build
+test/
+examples/
+*/build/
+*/*/build/
```
mnordsletten: You can do this for directories in dockerignore: `**/build` I think. See: https://docs.docker.com/engine/reference/builder/#dockerignore-file

Dockerfile
```diff
@@ -1,31 +1,110 @@
...
+    chmod 4755 /usr/local/bin/fixuid && \
+    mkdir -p /etc/fixuid && \
+    printf "user: $USER\ngroup: $GROUP\npaths:\n  - /service\n" > /etc/fixuid/config.yml
 
-USER ubuntu
+RUN echo "LANG=C.UTF-8" > /etc/default/locale
 
-ADD . /home/ubuntu/IncludeOS
-WORKDIR /home/ubuntu/IncludeOS
+# Docker TAG can be specified when building with --build-arg TAG=..., this is redeclared in the source-build stage
```
mnordsletten: Remove all the `BRANCH` and `REPO` references. That is no longer in use. (remember the comment too)

Dockerfile
```diff
@@ -1,31 +1,110 @@
...
+ENV BRANCH=$BRANCH
+ENV REPO=$REPO
+ENV NEWTAG=$NEWTAG
 
-RUN sudo apt-get update && \
-    sudo do_bridge="" ./etc/install_all_source.sh \
-&& sudo rm -rf /var/lib/apt/lists/*
 
-VOLUME /service
+LABEL dockerfile.version=1 \
```
mnordsletten: This LABEL can be removed too.

Dockerfile
```diff
@@ -1,31 +1,110 @@
...
+FROM base as source-build
+
+RUN apt-get update && apt-get -y install \
+    git \
+    lsb-release \
+    net-tools \
+    wget \
+    && rm -rf /var/lib/apt/lists/*
+
+
```
mnordsletten: Remove newlines and the echo below

Dockerfile
```diff
@@ -1,31 +1,110 @@
...
+# Copy IncludeOS contents from host repo to container
+RUN echo "copying Contents of current IncludeOS branch"
+RUN cd ~ && pwd && \
+  mkdir -p IncludeOS
+COPY . /root/IncludeOS/
+
+# Adding Custom git tags
+RUN echo "Assigning Your custom git tag"
```
mnordsletten: Either print what you sent in as custom tag, or remove the print.
Also there should not be a default tag. If no tag was sent in then it should just keep the same tag as was there from before. (make sure this works even if we are on the latest tag)

Dockerfile
```diff
@@ -1,31 +1,110 @@
...
+    apt-get remove -y python-pip && \
+    apt autoremove -y
+
+COPY --from=source-build  /usr/local/includeos /usr/local/includeos/
+COPY --from=source-build  /usr/local/bin/boot /usr/local/bin/boot
+COPY --from=source-build  /root/IncludeOS/etc/install_dependencies_linux.sh /
+COPY --from=source-build  /root/IncludeOS/etc/use_clang_version.sh /
+COPY --from=source-build  /root/IncludeOS/lib/uplink/starbase /root/IncludeOS/lib/uplink/starbase/
+COPY --from=source-build  /ios_version.txt /
+COPY --from=source-build  /root/IncludeOS/etc/docker_entrypoint.sh /entrypoint.sh
```
mnordsletten: Are we missing the `etc/docker_entrypoint.sh` file?

staiyeba: yes, will add it in the next PR

PR #1905, opened 2018-09-24
net: enable ENABLE_BUFFERSTORE_CHAIN toggling via CMake option

ingve:


ingve:
OBE (If I understand correctly, bufferstore has been rewritten and chained bufferstores are going away), so ignore this :)

PR #1907, opened 2018-09-24
net: BufferStore now uses internal pools instead of a chain

fwsGonzo:
- Greatly simplifies bufferstore
- Chaining removed, replaced with internal pools that can grow
- Growth can be disabled by overriding a private function


alfred-bratterud:
@fwsGonzo I like it. I was thinking statically allocating a fixed set of buffers would be nice as well (e.g. use buffer limit as a parameter to bufstore), but I see the advantages of growing dynamically as well. We can always add more options later. 

One thing we must have though is a stat counting all packets dropped due to sendq_limit reached, and also for when you can't refill receive ring due to buffer limit reached. With that I'll take it.

PR #1910, opened 2018-09-25
Support multiple bridges with qemu + simplification of Nic by MAC

AndreasAakesson:
With this small change, `vm.json` now supports the `scripts` member in `net`, which means qemu can use another ifup/ifdown script than the default one (`/includeos/scripts/qemu-ifup`).

The `create_bridge.sh` already (kinda) supported creating your own custom bridge, but now you also need to assign a mac address so you can have them different.

Usage:
`$ ./create_bridge.sh bridge44 10.0.1.1 255.255.255.0 c0:ff:0a:00:00:01`

Then I copy the `qemu-if*` scripts to `~/dev/bridge44/`, and edit the `BRIDGE` value inside the scripts to `bridge44`. Executable permission also needs to be set:
`$ chmod +x ~/dev/bridge44/qemu-if*`

In `vm.json` I point to the script path (they have to be named the same):
```
"net" : [
    {"device" : "vmxnet3", "mac" : "c0:01:0a:00:00:aa", "scripts": "/Users/andreas/dev/bridge44/"},
    {"device" : "vmxnet3", "mac" : "c0:01:0a:00:00:2a"}
  ]
```

This allows daredevils (like me) to create and try out the routing facilities we have. Not yet so stream lined - improvements are gladly welcomed!

### DIFF
--- a/api/hw/devices.hpp
+++ b/api/hw/devices.hpp
@@ -46,6 +46,12 @@ namespace hw {
     static Block_device& drive(const int N)
     { return get<Block_device>(N); }
 
+    // Nic helpers
+    inline static int nic_index(const MAC::Addr& mac);
+
+    static int nic_index(const std::string& mac)
+    { return nic_index(MAC::Addr{mac.c_str()}); }
+
     /** List all devices (decorated, as seen in boot output) */
     inline static void print_devices();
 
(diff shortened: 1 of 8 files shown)


AndreasAakesson:
Nic got helper functions to retrieve index for a given MAC addr. The result is that a stack with subindex (VLAN) can now also be retrieved together with MAC Addr.

PR #1912, opened 2018-09-25
Fix wrong calculation of heap end

alfred-bratterud:
Please review - turns out this is important


alfred-bratterud:
jenkins test please

PR #1925, opened 2018-10-04
Na cl find python fix

KristianJerpetjon:
Some flavors of Linux is moving to using python3 as default python hence executing python to get python2 fails this fix helps cmake locate the correct python interpreter it should be future proof for cmake 3.12+ which includes findpython2 and findpython3 packages


AnnikaH:
No errors here, @alfred-bratterud

PR #1926, opened 2018-10-04
Add memory read barrier to prevent speculative execution error

KristianJerpetjon:
Added a read ahead barrier to prevent desc.len from beeing read before its set.


AndreasAakesson:
jenkins test please

alfred-bratterud:
lgtm

PR #1933, opened 2018-10-05
Updating NaCl (iface and vlan functionality)

AnnikaH:
Includes https://github.com/includeos/NaCl/pull/62, https://github.com/includeos/NaCl/pull/63 and https://github.com/includeos/NaCl/pull/64


alfred-bratterud:
This was all green on Jenkins so the failure report is a false negative due to expermiental jenkins features

PR #1934, opened 2018-10-05
Prevent infinite CPU exception loops

alfred-bratterud:


alfred-bratterud:
I replaced the panic counter with a separate counter for cpu exceptions in order to avoid infinite cpu exception loops, which apparently can happen if something is severely broken. E.g. if trying to print registers causes a CPU exception. I disabled setting `is_panicking` from in here, which might not be the right thing - or if we keep this we might want to decrease the max number of panics the counter can reach before triggering double fault.

PR #1935, opened 2018-10-09
HTTP Client follow redirection

AndreasAakesson:
Default is 0 (can be changed static for *all* clients), or passed as an option.


staiyeba:
jenkins test please

PR #1936, opened 2018-10-09
run: some fixes to vmrunner

KristianJerpetjon:
Removed the dependency on python-magic as its different on different flavors of Linux causing unnecessary confusion and frustration 
 
file should however work on all flavors mac and Linux at a small execution cost
A nicer yet more comprehensive fix would be to wrap the c library libmagic manually using ctags 

Also removed --preserve-env as this feature depends on the sudo/selinux configuration this needs to be tested but bugsquash runs in qemu kvm without it. 

if we are keeping --preserve-env we need to test it and notify the user why and what to do as it does not fail gracefully

### DIFF
--- a/vmrunner/vmrunner.py
+++ b/vmrunner/vmrunner.py
@@ -10,7 +10,6 @@ import traceback
 import validate_vm
 import signal
 import psutil
-import magic
 from shutil import copyfile
 
 from prettify import color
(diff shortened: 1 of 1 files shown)


KristianJerpetjon:
jenkins test please

alfred-bratterud:
This works on my mac. @mnordsletten do you know why `--preserve-env` was added in the first place? Was hard to find in git history

Issue #1931, opened 2018-10-04
test.py's using subprocess directly needs their own timeouts


alfred-bratterud:
`vmrunner` has a timeout thread that will kill qemu and all child processes after a given time. This doesn't kill subprocesses started directly by `test.py`, such as `stress/test.py` and many others. The stress test is e.g. calling `subprocess.check_call(["httperf","--hog", ...)` etc. which apparently can hang forever.

One solution is to make the vmrunner timeout kill everything including itself, but that seems a little wrong wrt. what the vmrunner is supposed to do (run a vm, possibly for a finite time). Another solution is to require each test.py to be in charge of their own subprocesses. The subprocess module in Python 3 has a timeout parameter, which is also backported to python 2.7. 
`$ pip install subprocess32`
And then in `test.py`:
```
import subprocess32
...
supprocess32.check_call(..., timeout=5)
```
works for me, e.g. a call to `supprocess32.check_call(["yes"], timeout=10)` will result in an exception and `TimeoutExpired: Command '['yes']' timed out after 10 seconds` before control is handed back to `test.py` for graceful exit.

PR #1938, opened 2018-10-10
test: changes in process threads- Adding subprocess32 and thread_timeout

staiyeba:
#1931

Adding subprocess32 and thread_timeout where `subprocess_calls` seemed large and required a timeout that matches the test timeout.

A `thread_timeout` parameter has been added to assign a timeout for each test and use for two purposes so far.

For example,

-  the `boot(50)` where the `50` was the timeout has been replaced by `boot(thread_timeout)` 

- the subprocess call has been edited to use subprocess32 with a timeout as follows:

subprocess.check_output(["sudo", "hping3", "10.0.0.45", "-d", "20", "-0", "--ipproto", "16", "-c", "1"])

replaced with:
 
`subprocess32.check_output(["sudo", "hping3", "10.0.0.45", "-d", "20", "-0", "--ipproto", "16", "-c", "1"], timeout=thread_timeout)`

Thirteen (13) `test.py` files have been modified:   

**fs/integration/:**
fat32/test.py
ide/test.py
ide_write/test.py
vfs/test.py
virtio_block/test.py

**net/integration/:**
dhclient/test.py
dhcpd/test.py
dhcpd_dhclient_linux/test.py
dns/test.py
icmp/test.py
**stress/** test.py (had its own timeout parameter called `timeout `which is calculated in the script depending on the `BURST_COUNT`, the timeout parameter has been changed to `thread_timeout` to match all the other tests and also to avoid confusion with` subprocess32`’s own `timeout` parameter.

Added thread_timeout as a variable to two files:
**net/integration/router/** test.py
**posix/integration/syslog_plugin/** test.py

### DIFF
--- a/test/fs/integration/fat32/test.py
+++ b/test/fs/integration/fat32/test.py
@@ -2,6 +2,9 @@
 import sys
 import os
 import subprocess
+import subprocess32
+
+thread_timeout = 30
 
 includeos_src = os.environ.get('INCLUDEOS_SRC',
                                os.path.realpath(os.path.join(os.getcwd(), os.path.dirname(__file__))).split('/test')[0])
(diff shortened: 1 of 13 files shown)


alfred-bratterud:
@staiyeba the changes to test.py looks fine.  Keep in mind that`subprocess32` is not going to be present unless it's pip-installed. Right now we're doing `pip` in `install_dependencies_*.sh`, but since this change is only required for running tests we might just let it fail if it's not installed, as long as the jenkins machines get this dependency added automatically some other way.

staiyeba:
yes, will add subprocess32 into the dependencies file.

staiyeba:
jenkins test please

PR #1939, opened 2018-10-10
Guggg

fwsGonzo:


KristianJerpetjon:
lgtm if it builds :)

staiyeba:
jenkins test please

alfred-bratterud:
Thanks, we need this. With this however stats will not be persistent by default- I think they need to be. Maybe a quick fix is just to add a third ctor parameter Flags that defaults to persist. Flags would need to be an enum though - if you don't like enum class use old enum. Another flag could be`constant` - somebody asked for that to e.g. be able to expose rx/tx-ring sizes.

PR #1944, opened 2018-10-11
Extended gateway test for VLAN

AndreasAakesson:
Currently not run due to vmxnet3 with VLAN not working in qemu (being worked on).


AndreasAakesson:
Will fail until #1945 is merged.

Issue #1941, opened 2018-10-11
Rename/change Super_stack API.


AndreasAakesson:
`Super_stack` is not a good name. We've discussed renaming it to `net::Interfaces`. It currently also has a dependency with `net::Inet` that goes both way (for now good reason).

Renaming this should also include removing the current "super stack" dependency in Inet, and have "super stack" include Inet (to make everyones life simple).

What's causing the dependency now is:
```
    /** Return the stack on the given Nic */
    template <int N = 0>
    static auto&& Inet::stack()
    {
      return Super_stack::get(N);
    }
```
which is totally unnecessary. 
I suggest making the "super stack" the only endpoint to retrieve an Inet, by removing the above function.

PR #1947, opened 2018-10-12
Refactor Super_stack to Interfaces

AndreasAakesson:
Solve issue #1941.

NaCl need to be updated as well.

Probably a lot of failing tests using the old API (`Inet::stack<0>()`).


alfred-bratterud:
Nice!

PR #1948, opened 2018-10-12
test: fixing dhcpd_dhclient_linux timer error created earlier

staiyeba:
An `UboundLocalError` error was created due to adding `thread_timeout` in some subprocess calls (unnecessary use of subprocess32 call).
Removing the thread_timeout fixed the error:
```
Traceback (most recent call last):
File "/home/ubuntu/IncludeOS/vmrunner/vmrunner.py", line 861, in trigger_event
res = func(line)
File "test.py", line 122, in run_dhclient
timer.cancel()
UnboundLocalError: local variable 'timer' referenced before assignment
```

### DIFF
--- a/test/net/integration/dhcpd_dhclient_linux/test.py
+++ b/test/net/integration/dhcpd_dhclient_linux/test.py
@@ -96,29 +96,23 @@ def run_dhclient(trigger_line):
   print color.INFO("<Test.py>"), "Running dhclient"
 
   try:
-    dhclient = subprocess.Popen(
+    dhclient = subprocess32.check_output(
         ["sudo", "dhclient", "bridge43", "-4", "-n", "-v"],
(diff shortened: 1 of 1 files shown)


alfred-bratterud:
@staiyeba looks like Annika already added a timeout here - but I the subprocess32 timeout would allow us to simplify this test by not having to manage the timer ourselves. I'd suggest keeping the subprocess32 timeout and remove the old timer.

staiyeba:
It seems the `subprocess32.Popen` doesnot support a timeout. Therefore I have added the `thread_timeout` variable I created as a parameter to Python's threading module `timer` which Annika had already created. I suppose in the long run if we use python 3.5 we can use `subprocess.run` which supports a `timeout` parameter

alfred-bratterud:
Thanks @staiyeba

PR #1957, opened 2018-10-19
Various

fwsGonzo:


cmake/linux.service.cmake
```diff
@@ -136,6 +174,9 @@ endif()
 if (ENABLE_LTO OR USE_LLD)
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fuse-ld=lld")
 endif()
+if (LIBFUZZER)
```
KristianJerpetjon: option only for clang ?

cmake/openssl.cmake
```diff
@@ -2,36 +2,38 @@
 include(ExternalProject)
 
 if(${ARCH} STREQUAL "x86_64")
-  set(OPENSSL_HASH 3ef6bc4e8be049725ca3887f0d235031)
-
-  ExternalProject_Add(openssl_bundle
-          PREFIX openssl
-          URL https://github.com/fwsGonzo/OpenSSL_bundle/releases/download/v1.2/openssl_bundle.tar.gz
-          URL_HASH MD5=${OPENSSL_HASH}
+  ExternalProject_Add(s2n_bundle
```
KristianJerpetjon: shouldn't the openssl_bundle still be present or is it part of s2n bundle ?

lib/LiveUpdate/CMakeLists.txt
```diff
@@ -20,19 +12,25 @@ add_custom_command(
...
-if(${ARCH} STREQUAL "x86_64")
-  set(LIU_OPENSSL_FILES "serialize_openssl.cpp")
-endif()
-
 # LiveUpdate static library
 add_library(liveupdate STATIC
     storage.cpp partition.cpp update.cpp resume.cpp rollback.cpp
-    os.cpp serialize_tcp.cpp hotswap.cpp hotswap64_blob.asm
-    ${LIU_OPENSSL_FILES}
+    elfscan.cpp os.cpp "serialize_tcp.cpp"
```
KristianJerpetjon: why the "" ?

lib/LiveUpdate/partition.cpp
```diff
@@ -53,11 +54,12 @@ int storage_header::find_partition(const char* key) const
       // the partition must have a valid name
       assert(part.name[0] != 0);
       // the partition should be fully consistent
-      uint32_t chsum = part.generate_checksum(this->vla);
-      if (part.crc == chsum) {
-        return p;
+      if (LIVEUPDATE_USE_CHEKSUMS) {
```
KristianJerpetjon: Should we present a warning when this feature is off ?

src/util/uri.cpp
```diff
@@ -79,7 +80,9 @@ static inline util::sview updated_copy(const std::vector<char>& to_copy,
                                        util::csview view,
                                        const std::vector<char>& from_copy)
 {
-  return {to_copy.data() + (view.data() - from_copy.data()), view.size()};
+  // sometimes the source is empty, but we need a valid empty string
+  if (view.data() == nullptr) return {&to_copy.back(), 0};
```
KristianJerpetjon: next line with indent ?
avoids accidentally reading the next line as how the case is handled

test/kernel/integration/LiveUpdate/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ set(SOURCES
 # DRIVERS / PLUGINS:
 set(DRIVERS
     virtionet
-    boot_logger
+    #boot_logger
```
KristianJerpetjon: Remove entirely or will it come back later ? TODO ?

test/linux/liveupdate/test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+export CC=gcc-7
```
KristianJerpetjon: this should be a test for a GCC-7+

test/linux/s2n/test.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -e
+export CC=gcc-7
```
KristianJerpetjon: same comment as before make it test that gcc > 7 and that CC is GCC ?

api/net/s2n/stream.hpp
```diff
@@ -0,0 +1,338 @@
...
+    }
+  }
+  inline TLS_stream::~TLS_stream()
+  {
+    S2N_PRINT("s2n::TLS_stream::~TLS_stream(%p)\n", this);
+    assert(m_busy == false && "Cannot delete stream while in its call stack");
+    s2n_connection_free(this->m_conn);
+  }
+
+  inline void TLS_stream::write(buffer_t buffer)
```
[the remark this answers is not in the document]
fwsGonzo: Im writing a commit for this, clearly an error

cmake/openssl.cmake
fwsGonzo: S2N requires openssl 1.1, and I got tired of creating 2 bundles, so I made it into one. My s2n_bundle repo has all the deets.

src/util/uri.cpp
fwsGonzo: This bugfix is already on master, but we can change at least change the future. PRs welcome!

test/kernel/integration/LiveUpdate/CMakeLists.txt
fwsGonzo: It is probably meant to be always-on, as if it crashes it should print as much as possible.

PR #1975, opened 2018-11-12
conan: added conanfile for solo5

alfred-bratterud:


conan/solo5/0.3.1/conanfile.py
```diff
@@ -0,0 +1,28 @@
+import os
+from conans import ConanFile,tools,AutoToolsBuildEnvironment
```
KristianJerpetjon: you are not using the AutoToolsBuildEnvironment :)

alfred-bratterud: good point, will fix

PR #1976, opened 2018-11-12
Ip6: Improve SLAAC

AndreasAakesson:
* Supply token instead of alt addr.
  * May force to use token instead of eui64 generated from MAC.
* Configuring global address is now also on a timer.

DAD is still not working correctly (for link local). This is due to NDP RFC being long and boring.


KristianJerpetjon:
lgtm

AndreasAakesson:
IPv6 addr needs testing. I put the CASE in the wrong file due to the name of the ip6 addr test not having `_test`, but I'll fix this later together with getting unittest to build on macOS again.

PR #1979, opened 2018-11-13
S2N serialization work

fwsGonzo:
* Complex S2N serialization test with new config/context during "live updates"
* LiveUpdate support for streams (in general) and TLS streams, S2N TLS_stream being the first
* microLB changes:
    - Creating a load balancer now doesn't open it
    - Open for TCP: open_tcp(port)
    - Open for OpenSSL: open_ossl(port, cert_path, key_path)
    - Open for S2N: open_s2n(port, cert_path, key_path)

S2N needs to have a centralized initialization process in the OS so that we dont accidentally initialize it more than once. There is no test for live updating a microLB.


KristianJerpetjon:
lgtm but we should look into how to make certain features into options

PR #1986, opened 2018-11-18
ps2: Fix pressed-state for directional keys

fwsGonzo:
Extended scancodes also need to have their pressed-state checked


KristianJerpetjon:
lgtm

PR #1993, opened 2018-11-23
TCP_perf test: add option to keep uploaded file

alfred-bratterud:
I've added what I thought was the behavior of this test - it's useful for verifying consistency of data going in and out of an instance. You can now run this test with `boot . keep` to have the last file uploaded on port 1337 be the file you download when you cat from 1338. `boot . keep` looks weird, but apparently `--keep` would get the parameter eaten up by python argparse in boot. That's a separate issue.


examples/TCP_perf/service.cpp
```diff
@@ -113,15 +155,15 @@ void Service::ready()
...
+  blob = net::tcp::construct_buffer(SIZE, '!');
 
 #ifdef USERSPACE_LINUX
   extern void create_network_device(int N, const char* route, const char* ip);
   create_network_device(0, "10.0.0.0/24", "10.0.0.1");
 #endif
 
   // Get the first IP stack configured from config.json
-  auto& inet = net::Interfaces::get(0);
+  auto& inet = net::Super_stack::get(0);
```
AndreasAakesson: No idea how this snuck in here. And how it even passed on jenkins? This also goes for the include line 20.

PR #1994, opened 2018-11-24
AC97 and Intel HDA audio drivers

fwsGonzo:
- AC97 doesnt work because we dont have legacy interrupt (int#x) support
- Intel HDA is extremely complex, but the foundation is there

### DIFF
--- a/api/hw/audio_device.hpp
+++ b/api/hw/audio_device.hpp
@@ -1,0 +1,70 @@
+// This file is a part of the IncludeOS unikernel - www.includeos.org
+//
+// Copyright 2015-2017 Oslo and Akershus University College of Applied Sciences
+// and Alfred Bratterud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+#ifndef HW_AUDIO_DEVICE_HPP
+#define HW_AUDIO_DEVICE_HPP
+
+#include <cstdint>
+#include <delegate>
+#include <memory>
+#include <vector>
+
+namespace hw
+{
+  class Audio_device {
+  public:
+    enum mixer_value_t {
+      MASTER_VOLUME,
+      PCM_OUT
+    };
+    /**
+     * Method to get the type of device
+     *
+     * @return The type of device as a C-String
+     */
+    static const char* device_type() noexcept
+    { return "Audio device"; }
+
+    /**
+     * Method to get the name of the device
+     *
+     * @return The name of the device as a std::string
+     */
+    virtual std::string device_name() const = 0;
+
+    /**
+     * Get the human-readable name of this device's controller
+     *
+     * @return The human-readable name of this device's controller
+     */
+    virtual const char* driver_name() const noexcept = 0;
+
+    /* Get & set hardware mixer values */
+    virtual void     set_mixer(mixer_value_t, uint32_t) = 0;
+    virtual uint32_t get_mixer(mixer_value_t) = 0;
+
+    /**
+     * Method to deactivate the block device
+     */
+    virtual void deactivate() = 0;
+
+    virtual ~Audio_device() noexcept = default;
+  }; //< Audio_device
+} //< hw
+
+#endif
--- a/api/hw/msi.hpp
+++ b/api/hw/msi.hpp
@@ -42,7 +42,7 @@ namespace hw {
 
   struct msix_t
   {
-    msix_t(PCI_Device&, uint32_t capoff);
+    msix_t(PCI_Device&, bool msix, uint32_t capoff);
 
     // initialize msi-x tables for device
     void mask_entry(size_t);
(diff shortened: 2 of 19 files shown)


api/hw/audio_device.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Get the human-readable name of this device's controller
+     *
+     * @return The human-readable name of this device's controller
+     */
+    virtual const char* driver_name() const noexcept = 0;
+
+    /* Get & set hardware mixer values */
+    virtual void     set_mixer(mixer_value_t, uint32_t) = 0;
```
KristianJerpetjon: How do we get what mixer_value_t's are supported and how many

api/hw/msi.hpp
```diff
@@ -42,7 +42,7 @@ namespace hw {
 
   struct msix_t
   {
-    msix_t(PCI_Device&, uint32_t capoff);
+    msix_t(PCI_Device&, bool msix, uint32_t capoff);
```
KristianJerpetjon: Are we wrapping non msi_x interrupts thru msix ?

api/hw/ps2.hpp
```diff
@@ -79,11 +79,25 @@ namespace hw
     static keystate_t get_kbd_vkey();
     static uint8_t    get_mouse_irq();
 
+    // if you want to control PS/2 yourself
```
KristianJerpetjon: Exposing PS2 as a raw interface ? why not register the raw interface as an alternative driver ?

api/kernel/pci_manager.hpp
```diff
@@ -37,6 +38,9 @@ class PCI_manager {
   using BLK_driver = delegate< std::unique_ptr<hw::Block_device> (hw::PCI_Device&) >;
   static void register_blk(uint16_t, uint16_t, BLK_driver);
 
+  using SND_driver = delegate< std::unique_ptr<hw::Audio_device> (hw::PCI_Device&) >;
```
KristianJerpetjon: Hmm i wonder if we could do this in a more "generic" way to avoid having to add these in the header.

src/drivers/ac97.cpp
```diff
@@ -0,0 +1,138 @@
+#include "ac97.hpp"
```
KristianJerpetjon: Do we know if this works ?

src/drivers/ihda.cpp
```diff
@@ -0,0 +1,309 @@
+#include "ihda.hpp"
+#include <arch.hpp>
+#include <kernel/events.hpp>
+#include <immintrin.h>
+#include <info>
+
+#define REG_GCTL      0x08
```
KristianJerpetjon: change to defines to const ?

src/drivers/ihda.cpp
```diff
@@ -0,0 +1,309 @@
...
+      m_codecs.back().idx = i;
+    }
+  }
+}
+
+void IHDA::setup_ring(corb_t& ring, bool corb)
+{
+  assert(ring.reg != 0);
+  this->stop_dma(ring);
+  // we only support 256 buffers (mode=2)
```
KristianJerpetjon: typo in comment ? 
mode 2 is 16 buffers

src/drivers/ihda.cpp
```diff
@@ -0,0 +1,309 @@
...
+  uint8_t stat = this->read<uint8_t>(REG_STATESTS);
+  for (int i = 0; i < 16; i++) {
+    if (stat & (1 << i)) {
+      m_codecs.emplace_back();
+      m_codecs.back().idx = i;
+    }
+  }
+}
+
+void IHDA::setup_ring(corb_t& ring, bool corb)
```
KristianJerpetjon: Any point thinking about memory barriers ?

src/drivers/ihda.hpp
```diff
@@ -0,0 +1,95 @@
+#include <hw/audio_device.hpp>
```
KristianJerpetjon: Are some of these parts potentially reusable? Having private structures used in private functions in this header they all then perhaps should be a separate class used as private here.

src/hw/msi.cpp
```diff
@@ -113,8 +115,28 @@ namespace hw
     return pcibar.start + capbar_off;
   }
 
-  msix_t::msix_t(PCI_Device& device, uint32_t cap)
-    : dev(device)
+  msix_t::msix_t(PCI_Device& device, bool dev_msix, uint32_t cap)
```
KristianJerpetjon: Why do we not have a MSI alongside MSIX instead of wrapping the two together ?

api/hw/ps2.hpp
fwsGonzo: PS2 only has mouse and keyboard, there is no alternative :P

api/kernel/pci_manager.hpp
fwsGonzo: Alfred is doing the work now, this is the current dev, so Im not trying to reinvent here

src/drivers/ac97.cpp
fwsGonzo: Of course

src/drivers/ihda.cpp
fwsGonzo: I have tried to use them, especially one just before a branch that checks an MMIO variable, but it made no difference. Im not using those rings atm because the "immediate command" interface works better

src/hw/msi.cpp
fwsGonzo: We can move it out after it works, they do have some likeness

src/drivers/ihda.hpp
fwsGonzo: No, the API is quite strange, and I dont even think the streams carrying the audio data will share anything. Maybe the audio stream rings are more generic.

src/drivers/ihda.cpp
fwsGonzo: Indeed :+1:

PR #1995, opened 2018-11-25
Work towards making microLB test catch a nasty issue

fwsGonzo:
... which it does not, yet

### DIFF
--- a/api/net/stream.hpp
+++ b/api/net/stream.hpp
@@ -181,10 +181,23 @@ namespace net {
     **/
     virtual Stream* transport() noexcept = 0;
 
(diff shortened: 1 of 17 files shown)


fwsGonzo:
Apparently it does

alfred-bratterud:
Nice catch @fwsGonzo, thanks!

alfred-bratterud:
Thanks @fwsGonzo - I tested these changes now and they're working for me.

alfred-bratterud:
Looking to fix that LB test now, we'll merge when jenkins is green.

PR #1999, opened 2018-12-03
V0.13.x

alfred-bratterud:
WIP: Adjusting TCP buffer usage and receive window size


src/net/tcp/connection.cpp
```diff
@@ -649,6 +653,40 @@ void Connection::rtx_ack(const seq_t ack) {
   //  x-rtx_q.size(), rtx_q.size());
 }
 
+uint32_t Connection::calculate_rcv_wnd() const
```
KristianJerpetjon: Could be inlined ?

src/net/tcp/connection.cpp
```diff
@@ -688,21 +726,20 @@ void Connection::recv_data(const Packet_view& in)
   // this shouldn't be necessary with well behaved connections.
   // I also think we shouldn't reach this point due to State::check_seq checking
   // if we're inside the window. if packet is out of order tho we can change the RCV wnd (i think).
-  if(UNLIKELY(bufalloc->allocatable() < host_.max_bufsize())) {
+  /*if(UNLIKELY(bufalloc->allocatable() < host_.max_bufsize())) {
```
KristianJerpetjon: I :eyes: dead code

src/net/tcp/connection.cpp
```diff
@@ -936,12 +938,16 @@ void Connection::retransmit() {
...
-  else if(writeq.size()) {
+  else if(writeq.size())
+  {
     auto& buf = writeq.una();
+
+    // TODO: Finish to send window zero probe, but only on rtx timeout
+
     debug2("<Connection::retransmit> With data (wq.sz=%u) buf.unacked=%u\n",
-      writeq.size(), buf.length() - buf.acknowledged);
+           writeq.size(), buf->size(), buf->size() - writeq.acked());
```
KristianJerpetjon: Is it a :bird: ? is it a :airplane: ? 

no its a :taco:  :smile:

test/util/unit/pmr_alloc_test.cpp
```diff
@@ -14,11 +14,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//#define DEBUG_UNIT
+#define DEBUG_UNIT
 
 #include <common.cxx>
 #include <util/alloc_pmr.hpp>
 #include <util/units.hpp>
+
```
KristianJerpetjon: Have we tried this towards libcxx >= 7.0 ?

api/util/alloc_pmr.hpp
```diff
@@ -85,11 +86,16 @@ namespace os::mem {
     inline std::size_t dealloc_count();
     inline bool full();
     inline bool empty();
+
+    /** Fires when the resource has been full and is not full anymore **/
+    void on_non_full(Event e){ non_full = e; }
```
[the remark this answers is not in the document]
fwsGonzo: on_full_changed maybe?

src/net/tcp/connection.cpp
```diff
@@ -320,6 +321,7 @@ int  Connection::serialize_to(void*) const {  return 0;  }
 
 Packet_view_ptr Connection::create_outgoing_packet()
 {
+  update_rcv_wnd();
```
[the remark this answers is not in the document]
AndreasAakesson: It's an update on the receive window value for that Connection.

src/net/tcp/connection.cpp
AndreasAakesson: Inlining it means rebuilding half the OS when tampering with it. I wouldn't say the implementation is 100% yet.
Also, won't the compiler inline it for us inside?

src/net/tcp/connection_states.cpp
```diff
@@ -1061,6 +1070,7 @@ State::Result Connection::FinWait1::handle(Connection& tcp, Packet_view& in) {
     if(in.ack() == tcp.tcb().SND.NXT) {
       // TODO: I guess or FIN is ACK'ed..?
       tcp.set_state(TimeWait::instance());
```
AndreasAakesson: Maybe make a private helper function on Connection:
```
void enter_timewait()
{
  set_state(State::TimeWait::instance());
  release_memory();
  timewait_start();
}
```
and keep the `rtx_stop()` in those places that have it (order doesn't matter).
Also, the `if(tcp.rtx_timer.is_running())` is redundant since this check is already done in `Timer::stop()`.

src/net/tcp/tcp.cpp
```diff
@@ -509,15 +492,24 @@ bool TCP::unbind(const Socket& socket)
...
+bool TCP::add_connection(tcp::Connection_ptr conn) {
   // Stat increment number of incoming connections
   (*incoming_connections_)++;
 
   debug("<TCP::add_connection> Connection added %s \n", conn->to_string().c_str());
-  conn->_on_cleanup({this, &TCP::close_connection});
   conn->bufalloc = mempool_.get_resource();
+
+  // Reject connection if we can't allocate memory
+  if (conn->bufalloc == nullptr
```
AndreasAakesson: Maybe `UNLIKELY` here for extra cuteness (could probably be sprinkled on some more places where we check mem).

PR #2003, opened 2018-12-06
tcp: fix for checking that ack is within window

KristianJerpetjon:


alfred-bratterud:
Great, will test now!

PR #2008, opened 2018-12-11
Suggested code of conduct

perbu:
Hi,

A couple of months back we had a discussion about adding a CoC. The discussion didn't really conclude with anything so I thought I'll try to revive it and make sure we end up with something.  Since I'm somewhat familiar with the CoC adopted by Linux I picked it, but IMO anything will do. The main thing is that we have something in place before we need it.

I've put Alfred up as the arbitrator, a role I think he will do well. Hopefully, it won't be much work.

### DIFF
--- a/code-of-conduct.md
+++ b/code-of-conduct.md
@@ -1,0 +1,77 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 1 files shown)


alfred-bratterud:
Thanks @perbu

PR #2010, opened 2018-12-19
Building tests and examples in Conan branch

fwsGonzo:
While not everything will build due to some missing packages, there are still quite a few services that does.

### DIFF
--- a/conan/mana/conanfile.py
+++ b/conan/mana/conanfile.py
@@ -11,17 +11,15 @@ class ManaConan(ConanFile):
     generators = 'cmake'
     url = "http://www.includeos.org/"
 
-    #def build_requirements(self):
-        #eventually
-        #self.build_requires("includeos/%s@%s/%s"%(self.version,self.user,self.channel))
     def build_requirements(self):
+        self.build_requires("libcxx/{}@{}/{}".format("7.0", self.user,self.channel))
         self.build_requires("rapidjson/1.1.0@{}/{}".format(self.user,self.channel))
         self.build_requires("GSL/2.0.0@{}/{}".format(self.user,self.channel))
 
     def source(self):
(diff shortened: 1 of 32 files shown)


conan/mana/conanfile.py
```diff
@@ -11,17 +11,15 @@ class ManaConan(ConanFile):
     generators = 'cmake'
     url = "http://www.includeos.org/"
 
-    #def build_requirements(self):
-        #eventually
-        #self.build_requires("includeos/%s@%s/%s"%(self.version,self.user,self.channel))
     def build_requirements(self):
+        self.build_requires("libcxx/{}@{}/{}".format("7.0", self.user,self.channel))
```
KristianJerpetjon: you can write it as self.build_requires("libcxx/7.0@{}/{}".format(self.user,self.channel)
However the most correct would be to have self.build_requires("includeos/{}@{}/{}) as includeos provides the headers
Even more correct would be if we had a package for includeos_headers which we could depend on ..

conan/mana/conanfile.py
```diff
@@ -11,17 +11,15 @@ class ManaConan(ConanFile):
...
         self.build_requires("rapidjson/1.1.0@{}/{}".format(self.user,self.channel))
         self.build_requires("GSL/2.0.0@{}/{}".format(self.user,self.channel))
 
     def source(self):
-        #repo = tools.Git(folder="includeos")
-        #repo.clone("https://github.com/hioa-cs/IncludeOS.git",branch="conan")
-        shutil.copytree("/home/kristian/git/IncludeOS","IncludeOS")
+        #shutil.copytree($ENV{INCLUDEOS_SRC},"IncludeOS")
```
KristianJerpetjon: thanks.. :+1:

PR #2034, opened 2019-02-01
Fuzzery

fwsGonzo:
Updates fuzzing suite to work with clang-9, and fixes some IPv6 bugs


linux/userspace/CMakeLists.txt
```diff
@@ -102,6 +102,7 @@ set(OS_SOURCES
     ${IOS}/src/util/statman.cpp
     ${IOS}/src/util/path_to_regex.cpp
     ${IOS}/src/util/percent_encoding.cpp
+    #${IOS}/src/util/pmr_default.cpp
```
KristianJerpetjon: Typo ?

fwsGonzo: No, I need it for clang-9 so I need to be able to uncomment it

PR #2041, opened 2019-02-12
Conan: Updated profiles + added macos profile + Jenkinsfile

mnordsletten:
Now there are only `x86_64` profiles left.


mnordsletten:
Updated readme has been pushed. I also change to use the correct binutils recipe name.

PR #2042, opened 2019-02-13
Update chat link to point to slack

MagnusS:
Update README to link to slack instead of gitter.

Signed-off-by: Magnus Skjegstad <magnus@skjegstad.com>


ingve:
add to whitelist

PR #2078, opened 2019-03-13
Dockerfile: Simplified dockerfile that works for conan workflow

mnordsletten:
Does not cache any IncludeOS version, but inits conan with our config and contains the build tools (clang and nasm) required for building.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ RUN apt-get update
 RUN apt-get -y install \
     clang-6.0 \
     cmake \
+    nasm \
     curl \
     git \
     && rm -rf /var/lib/apt/lists/*
@@ -20,3 +21,9 @@ ENV CC=clang-6.0 \
     CXX=clang++-6.0
 RUN mkdir service
 WORKDIR service
+
+CMD mkdir -p build && \
+    cd build && \
+    conan install .. -pr clang-6.0-linux-x86_64 && \
+    cmake .. && \
+    make
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,31 @@
-from ubuntu:18.04
+FROM ubuntu:18.04
 
-RUN apt-get update
-RUN apt-get -y install \
-    clang-6.0 \
+ARG clang_version=6.0
+RUN apt-get update && \
+    apt-get -y install \
+    clang-$clang_version \
     cmake \
     nasm \
     curl \
-    git \
-    && rm -rf /var/lib/apt/lists/*
+    git && \
+    rm -rf /var/lib/apt/lists/*
 
 # Install and configure Conan
-ENV conan_version=1_12_3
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,111 +1,29 @@
...
-COPY --from=source-build  /usr/local/includeos /usr/local/includeos/
-COPY --from=source-build  /root/IncludeOS/etc/use_clang_version.sh /
-COPY --from=source-build  /root/IncludeOS/lib/uplink/starbase /root/IncludeOS/lib/uplink/starbase/
-COPY --from=source-build  /ios_version.txt /
-COPY --from=source-build  /version.txt /
-COPY --from=source-build  /root/IncludeOS/etc/docker_entrypoint.sh /entrypoint.sh
-ENTRYPOINT ["/entrypoint.sh"]
+# Configure build environment
+ENV CC=clang-6.0 \
+    CXX=clang++-6.0
```
MagnusS: Maybe these could be extracted from conan profile so we don't have to hardcode them here? E.g. `conan profile get env.CXX [profile name]`

Dockerfile
```diff
@@ -1,111 +1,29 @@
...
 CMD mkdir -p build && \
-  cd build && \
-  cp $(find /usr/local/includeos -name chainloader) /service/build/chainloader && \
-  echo "IncludeOS reported version:" $(cat /ios_version.txt) "label Version:" $(cat /version.txt) && \
-  cmake .. && \
-  make
+    cd build && \
+    conan install .. -pr clang-6.0-linux-x86_64 && \
+    cmake .. && \
+    make
```
ingve: Unpopular opinion: we should start using `cmake --build .` 😈 

[Build Tool Mode](https://cmake.org/cmake/help/v3.12/manual/cmake.1.html#build-tool-mode)

alfred-bratterud: @ingve nice - that makes sense for e.g. ninjas

Dockerfile
```diff
@@ -1,111 +1,29 @@
...
-COPY --from=source-build  /root/IncludeOS/lib/uplink/starbase /root/IncludeOS/lib/uplink/starbase/
-COPY --from=source-build  /ios_version.txt /
-COPY --from=source-build  /version.txt /
-COPY --from=source-build  /root/IncludeOS/etc/docker_entrypoint.sh /entrypoint.sh
-ENTRYPOINT ["/entrypoint.sh"]
+# Configure build environment
+ENV CC=clang-6.0 \
+    CXX=clang++-6.0
+RUN mkdir service
+WORKDIR service
```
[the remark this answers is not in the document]
mnordsletten: I now create the directory in a more dynamic way. If the user bind-mounts to a different dir they get the error:
```
ERROR: Conanfile not found at /service/conanfile.py or /service/conanfile.txt
```

PR #2080, opened 2019-03-13
Update solo5 to latest version

fwsGonzo:
... and make it possible to run hvt easier thanks to @niks3089


test_solo5_spt.sh
```diff
@@ -0,0 +1,17 @@
+#! /bin/bash
```
niks3089: May be we can move this test file and test_solo5_hvt.sh to test directory? These tests should be triggered with jenkins as well

vmrunner/vmrunner.py
```diff
@@ -328,6 +325,28 @@ def writeline(self, line):
...
+
+class solo5_spt(solo5):
+    def __init__(self, config):
+        super(solo5_spt, self).__init__(config)
+
+    def name(self):
+        return "Solo5-spt"
+
+    def boot(self, multiboot, debug=False, kernel_args = "", image_name = None):
+        solo5_bin = INCLUDEOS_HOME + "x86_64/bin/solo5-spt"
```
niks3089: There is a missing "/" before x86_64

test/misc/solo5-spt/test.sh
```diff
@@ -0,0 +1,177 @@
+#!/bin/bash
```
niks3089: This test also should be a part of jenkins to catch regressions

conan/solo5/0.4.1/conanfile.py
```diff
@@ -11,7 +11,7 @@ class Solo5Conan(ConanFile):
 
     def source(self):
         repo = tools.Git(folder = self.name)
-        repo.clone(self.url + ".git",branch="v{}".format(self.version))
+        repo.clone(self.url + ".git",branch="master")
```
niks3089: May be point to the latest solo5 commit instead by providing the commit id? We might not want their newer commits to break our tests

vmrunner/vmrunner.py
fwsGonzo: I thought I fixed that already.. let me check

fwsGonzo: Yes, its fixed in one of my commits

conan/solo5/0.4.1/conanfile.py
fwsGonzo: We don't actually use that path anymore.. The solo5 package is being built from another repo called "conan", so it would have to be changed there

vmrunner/vmrunner.py
niks3089: I thought that was fixed for hvt and not spt

etc/boot
```diff
@@ -157,14 +160,22 @@ if args.config:
...
     subprocess.call(['chmod', '+x', solo5_hvt])
     subprocess.call(['sudo', INCLUDEOS_PREFIX + "/scripts/solo5-ifup.sh" ])
 
+elif args.solo5_spt:
+    hyper_name = "solo5-spt"
+    os.environ['PLATFORM'] = "x86_solo5"
+    solo5_spt = INCLUDEOS_PREFIX + "/x86_64/bin/solo5-spt"
+
+    subprocess.call(['chmod', '+x', solo5_spt])
+    subprocess.call(['sudo', INCLUDEOS_PREFIX + "/includeos/scripts/solo5-ifup.sh" ])
```
niks3089: This should be "INCLUDEOS_PREFIX + "/scripts/solo5-ifup.sh" like above. The error comes up when running in rpr

fwsGonzo: Thanks, I've fixed that now

conan/solo5/0.4.1/conanfile.py
KristianJerpetjon: This file should live in the conan repo please update there. Do we depend on an unstable trunk ?  (replace with scm ?)

mnordsletten:
`boot` has also been moved to: https://github.com/includeos/vmrunner

PR #2084, opened 2019-03-14
Spring cleaning

AndreasAakesson:


fwsGonzo:
🛂

KristianJerpetjon:
We should do a script cleanup as well (etc be gone) + 2-3-4 sh files here and there should get better locations. if they should be executed with vm's they should be copied to vmrunner

PR #2089, opened 2019-03-19
editable: one step closer

KristianJerpetjon:
added layout.txt using jinja2 to parse the build folder correctly 
we can consider making multiple in the future.

using this code the following should work 

# Putting includeos into editable mode:
This is a bit of fiddling initially 
## Prepping the source 
git clone git@github.com:hioa-cs/IncludeOS.git
cd IncludeOS
comment out the version in the `conanfile.py`
`#version = get_version()`
(hopefully conan fixes that annoying issue)

## Do some local adoptions to where your build folder is
edit `layout.txt` (jinja2 syntax) and set your build folder in the first line
{% set build_dir='build' %}
[bindirs]
cmake
[includedirs]
api
[libdirs]
{{ build_dir }}/plugins
{{ build_dir }}/drivers
{{ build_dir }}/lib
[resdirs]
{{ build_dir }}

## set the conan package into editable mode
Make sure to adjust the version to whatever is apropriate (never than latest)
`conan editable add . includeos/0.15.0@includeos/test --layout=layout.txt`

## create a build folder and build includeos
Change this to whatever build folder matches the `layout.txt`
mkdir build 
cd build
conan install .. -pr <conan_profile> (-o options like solo5=True etc)
cmake .. (-DTOOLCHAIN_PROFILE=xx.yy when needed)
make

# Status
The package is now in editable mode and any dependencies should pick this package from your local cache. if you change anything to the code a simple make should be enough however if your dependencies change you need to redo the conan install step.

example on how it looks when its pulled into cache as editable    includeos/0.15.0@includeos/test:45955af12a7f7608830c1d255b80a75440406e3c - Editable

## finalizing the work / changes
Once the code is “finalized and if you want to verify that the conan package still builds 
remove the editable `conan editable remove includeos/0.15.0@includeos/test`
remove the comment on the #version and do a normal `conan create <source_path> includeos/test -pr <conan_profile>`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,48 @@
 cmake_minimum_required(VERSION 3.1.0)
 
+project (includeos C CXX)
+
+option(WITH_SOLO5 "Build with solo5 support" OFF)
+
+#Sets the includeos default profile to clang-5.0
+if (NOT DEFINED CONAN_PROFILE)
+  SET(CONANPROFILE "")
+else()
+  set(CONANPROFILE PROFILE ${CONAN_PROFILE})
+endif()
+
+#Are we executing cmake from conan or locally
+#if locally then pull the deps from conanfile.py
+#if buiding from conan expect the conanbuildinfo.cmake to already be present
+if(CONAN_EXPORTED OR EXISTS ${CMAKE_CURRENT_BINARY_DIR}/conanbuildinfo.cmake) # in conan local cache
+  # standard conan installation, deps will be defined in conanfile.py
+  # and not necessary to call conan again, conan is already running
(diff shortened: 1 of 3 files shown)


staiyeba:
> Can the `layout.txt` live in the `etc/` directory? I'm trying to clean up as much as possible in the root of the repo.
> Also it looks like you are using conan `0.13.0` to run some of these commands. Might be worth mentioning. We might need to decide if we are moving to `0.13.0` for Jenkins as well. Conan releases seem to be very frequent right now.
> 
> Other than that this looks great. I'll test this tomorrow.

Has this been tested? Are we going with conan release `1.13.1` ?

PR #2092, opened 2019-03-22
Preparing for release. Stable 3rd party dependencies + stable/latest uploads

mnordsletten:
Needed a bit of a trick to get the correct upload CHANNEL configured. 

All 3rd party dependencies will now use the stable channel. The remote has been changed in the conan_config repo. 

New behavior is:
- PRs are not uploaded anywhere (this is same as before)
- Merges to master/dev will get uploaded to `includeos/latest`.
- Tags/releases will upload to `includeos/stable`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,13 +14,34 @@ pipeline {
     CC = 'clang-6.0'
     CXX = 'clang++-6.0'
     USER = 'includeos'
-    CHAN = 'test'
+    CHAN = 'default'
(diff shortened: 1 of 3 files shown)


mnordsletten:
> S2N is not version 1.1.1 its a dummy to make things work but its not based on an official TAG to say its stable is imho a bit wrong. but i am open to discuss how to handle it

@KristianJerpetjon I've changed to the `0.8` package that was uploaded last friday. I've also added the `python_requires` functionality to this pr.

PR #2104, opened 2019-03-27
fix CMake out of source build failing due to incbin instruction

ilelann:
To fix this error I got when building out of source (Fedora 29,  Qt Creator 4.7.2):

/home/ilelann/src/IncludeOS/lib/LiveUpdate/hotswap64_blob.asm:21: error: `incbin': unable to get length of file `hotswap64.bin'

With this patch, you can just open CMakeLists.txt in Qt Creator and things seem to build fine.
Only assuming an override of CMAKE_INSTALL_PREFIX in CMake Configuration of used Qt Creator kit to work around the IDE trying to write to /usr/local .

Disclaimer:
I am totally new to IncludeOS. Currently playing with it. 
So this is not tested and I have no idea if this breaks the whole universe. :)

### DIFF
--- a/lib/LiveUpdate/CMakeLists.txt
+++ b/lib/LiveUpdate/CMakeLists.txt
@@ -30,6 +30,10 @@ add_library(liveupdate STATIC
     os.cpp serialize_tcp.cpp hotswap.cpp hotswap64_blob.asm
     ${LIU_OPENSSL_FILES}
   )
+
+# to allow out of source build, even if hotswap64_blob.asm contains incbin "hotswap64.bin"
+set_source_files_properties(hotswap64_blob.asm PROPERTIES COMPILE_FLAGS "-I ${CMAKE_CURRENT_BINARY_DIR}/")
+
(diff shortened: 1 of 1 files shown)


ilelann:
wrongly create a PR to master, will create one to dev branch...

ilelann:
Having unrelated local issues with dev branch (conan setup mostly) ,  I switch back the PR to master..
Waiting for comments.

KristianJerpetjon:
This change probably makes current master work in QT for you now, but i am unsure if everything is buildt out of tree. (I did fix some of it before moving to conan). 

I have fixed afaik every out of tree build issue on dev but it's untested 

current dev doesn't need install location and unless you are changing the includeos source you dont even have to build the includeos binary. Building the examples from the demo-examples repo should be possible without installing includeos. (but you need conan) please join us on slack https://goo.gl/NXBVsc to discuss issues

mnordsletten:
We will close this PR as it has been stale for over a month. Also when we merge the dev branch to master it will be outdated.

PR #2107, opened 2019-03-28
Fix build type debug

alfred-bratterud:


mnordsletten:
This PR needs to fix the instabilities in `kernel_LiveUpdate` test before merge.

PR #2124, opened 2019-04-04
UDPv6 test case

AndreasAakesson:


AndreasAakesson:
> Great that you got it to work. What was the issue yesterday?

What a struggled with yesterday was to correctly use `getaddrinfo()` which seems to be necessary for most of the cases when you wanna use a IPv6 address with sockets.

PR #2141, opened 2019-04-11
NaCl: Fix NaCl support

mnordsletten:
Still has python2 :( Will create issue to remove this at a later date for nacl.
I'll also create an issue to extract all the NaCl cmake code.

### DIFF
--- a/cmake/os.cmake
+++ b/cmake/os.cmake
@@ -354,12 +354,18 @@ function(internal_os_add_config TARGET CONFIG_JSON)
   target_link_libraries(${TARGET}${TARGET_POSTFIX} --whole-archive config_json_${TARGET} --no-whole-archive)
 endfunction()
 
-#TODO fix nacl
 function(os_add_nacl TARGET FILENAME)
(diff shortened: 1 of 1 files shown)


cmake/os.cmake
```diff
@@ -354,12 +354,18 @@ function(internal_os_add_config TARGET CONFIG_JSON)
   target_link_libraries(${TARGET}${TARGET_POSTFIX} --whole-archive config_json_${TARGET} --no-whole-archive)
 endfunction()
 
-#TODO fix nacl
 function(os_add_nacl TARGET FILENAME)
-  set(NACL_PATH ${CONAN_RES_DIRS_INCLUDEOS}/tools/NaCl)
+  find_program(PYTHON2_EXECUTABLE python)
+  if (PYTHON2_EXECUTABLE-NOTFOUND)
+    message(FATAL_ERROR "Python not found")
```
KristianJerpetjon: what is the minimum required cmake version for this ?

mnordsletten: I'm not a 100% sure, but `3.0.2` is the earliest mention in the docs. Anyway, I should be using the same functionality as the rest of the `os.cmake`

PR #2142, opened 2019-04-14
More robust heap initialization w/LiveUpdate and SystemLog

fwsGonzo:


src/hal/machine.cpp
```diff
@@ -89,6 +89,18 @@ namespace os::detail {
     MINFO("Initializing heap\n");
     auto main_mem = memory().allocate_largest();
     MINFO("Main memory detected as %zu b\n", main_mem.size);
+    memory().deallocate(main_mem.ptr, main_mem.size);
+
+#ifndef PLATFORM_x86_solo5
```
KristianJerpetjon: We could use something more generic from cmake ie PLATFORM_SUPPORTS_LIVEUPDATE ?

src/hal/machine.cpp
```diff
@@ -89,6 +89,18 @@ namespace os::detail {
...
+    static const size_t SYSTEMLOG_SIZE = 65536;
+    const size_t LIU_SIZE = (main_mem.size / (100 / 25)) & ~0xfff;
```
KristianJerpetjon: the last part is to make sure its rounded down to 4k ? so you get somewhat less than 25% of all ram

lib/LiveUpdate/src/os.cpp
```diff
@@ -3,47 +3,43 @@
...
-  return phys_max / (100 / size_t(LIVEUPDATE_AREA_SIZE));
-}
-uintptr_t kernel::liveupdate_phys_loc(size_t phys_max)  noexcept {
-  return phys_max & ~(uintptr_t) 0xFFF;
-}
-
 void kernel::setup_liveupdate(uintptr_t phys)
 {
-  static uintptr_t temp_phys = 0;
+  static uintptr_t temp_phys;
```
alfred-bratterud: Explicitly removing zero-initialization. Is there a better way to solve this?

src/hal/machine.cpp
```diff
@@ -89,6 +89,18 @@ namespace os::detail {
     MINFO("Initializing heap\n");
     auto main_mem = memory().allocate_largest();
     MINFO("Main memory detected as %zu b\n", main_mem.size);
+    memory().deallocate(main_mem.ptr, main_mem.size);
```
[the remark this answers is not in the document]
fwsGonzo: That was essentially what you did before, I just moved it

PR #2145, opened 2019-04-23
openssl: Fix stream busy depth bug

fwsGonzo:
This bug tripped up the encrypted websockets benchmark


src/net/openssl/tls_stream.cpp
```diff
@@ -145,9 +149,9 @@ void TLS_stream::handle_read_congestion()
 {
   //Ordering could be different
   send_decrypted(); //decrypt any incomplete
-  this->m_busy=true;
+  this->m_busy += 1;
```
KristianJerpetjon: The +1 -1 could be replaced by a RAII pattern like lock guard..

KristianJerpetjon:
Lets re visit the +1-1 pattern someday

PR #2150, opened 2019-04-29
threads: Add support for pthread_create/phtread_exit

fwsGonzo:
Support for the most important parts of the pthread API. No paralellization support.


fwsGonzo:
This will need some stubbing to get it to play nice when building the aarch64 platform

PR #2155, opened 2019-04-30
test: Reduce far_distance to 1 GiB in the paging test to avoid crash …

fwsGonzo:
…with LU area

@mnordsletten


mnordsletten:
Please update the vmrunner version as well. Then it will use correct CPU flags.

mnordsletten:
So `net_udp` test can be fixed by giving it more ram (again).
The liveUpdate test is still very unstable when I try to run it.

PR #2160, opened 2019-05-03
travis-ci: build for linux with clang-6.0 x86_64 on macos

staiyeba:
Building same version of IncludeOS kernel and LiveUpdate for Linux on **MacOS**:
profile: `clang-6.0-macos-x86_64`
arch: `x86_64`
compiler: `clang-6.0`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,9 @@ before_cache:
   - brew cleanup
 
 cache:
-  ccache: true
   - directories:
     - $HOME/Library/Caches/Homebrew
-    
+
 jobs:
   include:
     - script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,32 @@
+os: osx
+osx_image: xcode10.2
+compiler:
+  - clang
+  - gcc
+
+addons:
+  homebrew:
+    packages:
+    - cmake
+    - conan
+    - nasm
+    - llvm@6
+
+jobs:
+  include:
+    - script:
+        - export CC=/usr/local/opt/llvm\@6/bin/clang
+        - export CXX=/usr/local/opt/llvm\@6/bin/clang++
+        - echo "BUILDING KERNEL"
+        - conan create . includeos/latest -pr clang-6.0-macos-x86_64
+        - echo "BUILDING LiveUpdate"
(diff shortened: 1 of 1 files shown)


staiyeba:
Can someone review and see if we want to test building on mac for any other compiler and/or architecture.
From my observation I think any other one will require a profile to be written first.
However if this is all, this PR is running here atm: https://travis-ci.org/staiyeba/IncludeOS/builds/527685318

staiyeba:
Adding another stage for a different specification requires the following template:

```
   - script:
        - echo "lets do gcc for arch"
        - brew install gcc@8
      name: "gcc-8.2-aarch64"
```

.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+    packages:
+    - cmake
+    - conan
+    - nasm
+    - llvm@6
+
+jobs:
+  include:
+    - script:
+        - export CC=/usr/local/opt/llvm\@6/bin/clang
```
mnordsletten: We shouldn't need to export CC and CXX when using a conan profile. Have you tried without this?

staiyeba: I will re-check that, there was some confusion with the compiler path ... but it's probably fixed now.
Will write back after verification.

staiyeba:
At the moment, it takes about 13minutes without the cache.
My estimate with a cache, it will take about 3 minutes less.
Adding a boot and run step will thus add up to 13 minutes.

The other recommended feature is to add a downstream trigger, which has been a bit messier then I expected, thus might not be able to provide that right away as it's not a priority atm.

staiyeba:
As for the compiler path,
we have only the name of the binary as follows:
```
CC=clang-6.0
CXX=clang++-6.0
```
but this doesnot seem to be able to provide the path correctly, thus I have had to initialize them in the environment as follows: 
```
CC=/usr/local/opt/llvm\@6/bin/clang
CXX=/usr/local/opt/llvm\@6/bin/clang++
´´´

If we fix this in the conan_config repo and it works for everyone, then they do not need to be initialised in the environments here.

staiyeba:
Actually I rechecked the compiler path and it is correct, however macos on travis can't seem to use the path through the profile, so add it on travis job env settings

PR #2172, opened 2019-05-27
Travis CI Pipeline to Test Build Step

staiyeba:
1. Deploys a MacOS Instance
2. Installs necessary tools
3. Pulls latest master and fetches latest tags
4. Builds:
   - IncludeOS kernel
   - LiveUpdate
5. Caches `$HOME/Library/Caches/Homebrew`
   - Note: A cleanup of brew is carried out before caching, this ensures that the cache doesn't get bulky with unnecessary things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,38 @@
+os: osx
+osx_image: xcode10.2
+compiler:
+  - clang
+  - gcc
+
+addons:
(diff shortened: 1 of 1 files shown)


MagnusS:
Closing this PR for now. We can resurrect it if we need travis/osx support

PR #2220, opened 2023-03-12
Update README for themewise logo

IMPranshu:
The readme file has been updated to select the logo based on the system's theme preference as the original logo was not readable in a dark theme.


MagnusS:
Thanks for the contribution, but we'd like to avoid merging contributed binaries into the repo and this is just for a cosmetic change. I'm closing this for now - perhaps we can add support for this later with a binary produced from the original logo material

PR #2226, opened 2024-04-30
nix: Use clang as stdenv

MagnusS:
Sets clang as stdenv and updates to 23.11 to get `clang6Stdenv`. I used Clang 6 for now as it matches the old conan profile. 

uzlib is also patched to use $(CC) instead of gcc. Later versions have fixed this, so when we can upgrade we will not need this patch.


fwsGonzo:
24.04 is just out, but perhaps that is too new?

PR #2228, opened 2024-05-01
Conan2

alfreb:
Adds conan2 packages for most dependencies. I haven't been able to get the whole thing to link yet, but I think we might as well get these in and iterate from there.


MagnusS:
This looks ok to merge as all changes are in `deps` for now -- getting conan2 to work first and then move everything gradually to nix seems reasonable

PR #2229, opened 2024-05-01
Remove nix patches

alfreb:
Apply the patches from the nix file and remove them. Also install service_name.cpp so we can remove the dependency on the source tree for building a service.


MagnusS:
Tested locally with nix-build - lgtm

PR #2230, opened 2024-05-01
Build musl for IncludeOS with nix and clang

alfreb:
Should build a useable musl for includeos with `nix-build` from `IncludeOS/deps/musl`

### DIFF
--- a/deps/musl/default.nix
+++ b/deps/musl/default.nix
@@ -1,0 +1,49 @@
+{ nixpkgs ?
+  builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/refs/tags/23.11.tar.gz"
+, pkgs ? import nixpkgs { config = {}; overlays = []; }
+}:
+with pkgs;
+let
+  stdenv = clang7Stdenv;
+  musl-includeos=stdenv.mkDerivation rec {
+    pname = "musl-includeos";
+    version = "1.1.18";
+
+    src = fetchGit {
+      url = "git://git.musl-libc.org/musl";
+      rev = "eb03bde2f24582874cb72b56c7811bf51da0c817";
+    };
+
+    nativeBuildInputs = [ git clang tree];
+
+    patches = [
+      ./patches/musl.patch
+      ./patches/endian.patch
+    ];
+
+    postUnpack = ''
+      echo "Replacing musl's syscall headers with IncludeOS syscalls"
+
+      cp ${./patches/includeos_syscalls.h} $sourceRoot/src/internal/includeos_syscalls.h
+      cp ${./patches/syscall.h} $sourceRoot/src/internal/syscall.h
+
+      rm $sourceRoot/arch/x86_64/syscall_arch.h
+      rm $sourceRoot/arch/i386/syscall_arch.h
+      '';
+
+   configurePhase = ''
+      echo "Configuring with musl's configure script"
+      ./configure --prefix=$out --disable-shared --enable-debug
+      '';
+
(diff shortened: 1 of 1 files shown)


deps/musl/default.nix
```diff
@@ -0,0 +1,49 @@
+{ nixpkgs ?
+  builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/refs/tags/23.11.tar.gz"
```
bjornfor, line 2: Tip: if you supply fetchTarball with `{ url = ...; sha256 = ...; }` instead of just a string, Nix will cache the download indefinitely. Currently it will cache for 1 hour or so, and then re-download next time `nix-build` is run.

deps/musl/default.nix
```diff
@@ -0,0 +1,49 @@
...
+, pkgs ? import nixpkgs { config = {}; overlays = []; }
+}:
+with pkgs;
+let
+  stdenv = clang7Stdenv;
+  musl-includeos=stdenv.mkDerivation rec {
```
bjornfor, line 8: Nit: missing space around `=`.

deps/musl/default.nix
```diff
@@ -0,0 +1,49 @@
...
+
+    postUnpack = ''
+      echo "Replacing musl's syscall headers with IncludeOS syscalls"
+
+      cp ${./patches/includeos_syscalls.h} $sourceRoot/src/internal/includeos_syscalls.h
+      cp ${./patches/syscall.h} $sourceRoot/src/internal/syscall.h
+
+      rm $sourceRoot/arch/x86_64/syscall_arch.h
+      rm $sourceRoot/arch/i386/syscall_arch.h
+      '';
```
bjornfor, line 32: Nit: The closing quotes are typically de-indented one level (2 spaces).

deps/musl/default.nix
```diff
@@ -0,0 +1,49 @@
...
+
+   configurePhase = ''
+      echo "Configuring with musl's configure script"
+      ./configure --prefix=$out --disable-shared --enable-debug
+      '';
```
bjornfor, line 37: Nit: The closing quotes are typically de-indented one level (2 spaces).

deps/musl/default.nix
```diff
@@ -0,0 +1,49 @@
...
+
+    CFLAGS = "-Wno-error=int-conversion -nostdinc";
+
+    meta = {
+      description = "musl - an implementation of the standard library for Linux-based systems";
+      homepage = "https://www.musl-libc.org/";
+      license = lib.licenses.mit;
+      maintainers = with lib.maintainers; [ includeos ];
```
bjornfor, line 45: AFAICT, `includeos` isn't defined anywhere, so if something/someone were to evaluate meta.maintainers, that is going to fail.

Nix is lazy; it won't evaluate things that isn't used and `nix-build` won't fail because it doesn't need to evaluate `meta.maintainers` to be able to build the package. (Most attributes that get passed to stdenv.mkDerivation do get evaluated, but `passthru` and `meta` are exceptions.)

deps/musl/default.nix
alfreb: Nice, fixed!  Also this helped me find `nix-prefetch-url --unpack` which gives me the right sha, plus it seems to install the tarball into the store, which is exactly what I would have hoped for.

deps/musl/default.nix
alfreb: Good catch - we should probably find a project wide way of doing this. I've deleted the field for now.

PR #2233, opened 2024-05-10
First steps towards nix static build

MagnusS:
Initial work to create a static nix build. Still has some errors and can't fully build/link includeos yet, but should be useful foundation. Should be followed up by (rebased) #2231 

cc @alfreb


alfreb:
Awesome, thanks!

PR #2234, opened 2024-05-11
Reintegrate vmbuild / elf_syms as includeos-tools

alfreb:
- Some git history still lives in includeos/vmbuild, but most is in older versions of IncludeOS. See README.md for details.
- Removes the need for custom built cmake in default.nix by being more explicit about when we need pkgs and pkgs.pkgsStatic.

NOTE: To test the build, use `IncludeOS $ nix-build -A tools` . I added them to the toplevel nix, but since they are not a build dependency for IncludeOS they won't build with just `nix-build`. 

@fwsGonzo please comment on this- elf_syms is your work. Happy to remove elf_syms from this PR and let you reintegrate that yourself if you prefer.


fwsGonzo:
If the history is lost you can add me as a committer on the files I wrote, and also as an author on the top of files :) 
But, how did the history get lost?

alfreb:
> If the history is lost you can add me as a committer on the files I wrote, and also as an author on the top of files :) But, how did the history get lost?

The history isn't really lost - it's just not visible on these files anymore since I've copied them back in from here: 
https://github.com/includeos/vmbuild

They were moved out. But as the readme shows the old history is still here.

alfreb:
I'll redo this later.

PR #2237, opened 2024-05-12
It links! An example binary linking with IncludeOS (WIP)

alfreb:
It links, but we don't know if it boots yet. And it's not entirely clear if it will work since `musl-includeos` is used for linking, but musl from pkgsStatic is used for compiling. But it does contain the IncludeOS symbols such as `rock_bottom` and `_start` at the right order, and is definitely using our linker script. So it's progress. 

Note that we may want to revisit how this is done based on the discussion #2236.

### DIFF
--- a/cmake/os.cmake
+++ b/cmake/os.cmake
@@ -6,19 +6,13 @@ set (CMAKE_CXX_STANDARD 17)
 set (CMAKE_CXX_STANDARD_REQUIRED ON)
 
 
-# Helper to assert that env vars are set and convert them to local vars.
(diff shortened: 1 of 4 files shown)


MagnusS:
I suggest we merge this before #2236, as it brings us further towards a working build. It's then also easier to test that the refactoring in #2236 gives us the same result.

It would be nice with a more descriptive commit message though ;-)

alfreb:
Updated commit message with all the things

MagnusS:
Thanks! LGTM

PR #2238, opened 2024-05-16
Separate nix-file for example and vmbuild, fix conflicts

MagnusS:
This adds vmbuild.nix to build the vmbuild-tools and moves the example from default.nix to example.nix.

They can be built separately with:

nix-build ./vmbuild.nix
nix-build ./example.nix

This should also remove the conflicts from #2236

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -10,10 +10,8 @@
 { nixpkgs ? ./pinned.nix, # Builds cleanly May 9. 2024
   pkgs ? (import nixpkgs { }).pkgsStatic,
 
-  llvmPkgs ? pkgs.llvmPackages_16,
-
   # This env has musl and LLVM's libc++ as static libraries.
-  stdenv ? llvmPkgs.libcxxStdenv
+  stdenv ? pkgs.llvmPackages_16.libcxxStdenv
 }:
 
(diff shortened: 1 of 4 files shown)


MagnusS:
After #2236 is merged, we could hopefully also move the current default.nix to includeos.nix and make a toplevel nix-file that builds everything - but I haven't tested this yet.

alfreb (approved):
Looks really good! (and builds cleanly with `nix-build example.nix` / `nix-build vmbuild.nix`). I like that they are decoupled and that I can do this:

```
~/IncludeOS$ nix-build
/nix/store/12fvih0pm1g98r7z3vikh74f9h2ww6y1-includeos-static-x86_64-unknown-linux-musl-dev
```
to get my IncludeOS package and not just the example binary.

I think we'll want the example to also depend on `vmbuild` but right now it doesn't so we can hook that up when needed. It's a TODO in `cmake/os.cmake` to hook elf_syms back in - not sure if we want to do it in cmake or nix actually - I think I prefer nix; it reduces dependencies between cmake files and nix is a nicer language.

MagnusS:
Thanks - I'll merge this to unblock https://github.com/includeos/IncludeOS/pull/2236

PR #2239, opened 2024-05-19
Working nix-shell - a bit convoluted but allows quick iteration.

alfreb:
I've been trying to make a nix shell that can take an IncludeOS package directly from a cmake install as input, to allow fast iteration on OS features, that link into a bootable, without having to rebuild the OS every time.  

- `shell.nix` is working to the point where cmake configure step finds the right versions of libcxx, libcxxabi, liunwind and musl, but compilation doesn't pick up the right clang by default. I hack it in by defining $CXX, but then it doesn't find libcxx, so c++ includes like `<string>` are unavailable. 

- `shell_overlay.nix` is a little simpler, where the idea was to expose the paths to the link time dependencies from `pkgsIncludeOS`. It has the same result - I get to the point where I think I can link, but I can't compile the bootable binary due to missing includes.

How can we make the nix shell use the same stdenv as `nix-build`?

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,67 @@
+{ nixpkgs ? (import ./pinned.nix { }),
+  includeos ? import ./default.nix { },
+  pkgs ? nixpkgs.pkgsStatic,
+  llvmPkgs ? pkgs.llvmPackages_16
+}:
+pkgs.mkShell rec {
+
+  stdenv = pkgs.llvmPackages_16.libcxxStdenv;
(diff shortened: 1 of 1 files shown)


MagnusS:
Perhaps we could use `substituteInPlace` to replace the paths we need in the make files directly. See the solo5 example in  https://nixos.org/manual/nixpkgs/stable/#ssec-stdenv-dependencies-overview-example. If we substitute the paths to point to the right location in `cmake/os.cmake` during `postInstall` we wouldn't have to specify them everywhere.

PR #2240, opened 2024-05-20
Only include required files in IncludeOS nix package

MagnusS:
This avoids rebuilding the OS when working on in-tree packages that depend on `./default.nix`, such as `./example.nix`.

Also enables parallel building, but I'm not sure this affects build time at all.

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -28,9 +28,6 @@ final: prev: {
             ./deps
             ./userspace
             ./lib
-            ./default.nix
-            ./overlay.nix
-            ./pinned.nix
             ./CMakeLists.txt
           ];
       };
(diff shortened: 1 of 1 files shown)


overlay.nix
```diff
@@ -13,11 +13,27 @@ final: prev: {
...
+          root = ./.;
+          # Only include files needed by IncludeOS (not examples, docs etc)
+          fileset = prev.pkgsStatic.lib.fileset.unions [
+            ./src
+            ./api
+            ./cmake
+            ./deps
+            ./userspace
+            ./lib
+            ./default.nix
```
bjornfor, line 31: Nix files are typically only needed at eval time. You should be able to remove them from this list and in the future be able to refactor the Nix files without having to do any rebuilds (as long as the result evaluates to the same derivations).

overlay.nix
```diff
@@ -13,11 +13,27 @@ final: prev: {
...
+            ./overlay.nix
+            ./pinned.nix
```
MagnusS, line 33: Thanks, I'll remove them

```suggestion
```

PR #2246, opened 2024-06-08
Package unikernel dependencies with IncludeOS

alfreb:
- overlay.nix now copies libc, libcxx and friends to the IncludeOS package
- os.cmake is made responsible for finding libs inside the package
- delete library env vars from chainloader and example, no longer needed

Tested with `nix-build chainloader.nix && nix-build example.nix && boot result/bin/hello_includeos.elf.bin`


MagnusS:
Looks good!

PR #2249, opened 2024-06-10
Upgrade to C++20, delete GSL, fix some tests 

### LINKED ISSUE #2242: Move to C++20
We're currently on C++17 and use `span` etc from an [older version of GSL](https://github.com/includeos/IncludeOS/blob/v0.16.0-release/deps/GSL/default.nix). The old nixpkgs version doesn't build GSL on darwin (due to https://github.com/NixOS/nixpkgs/issues/217856).

Moving to C++20 should allow us to unpin GSL and use the latest version from nixpkgs, and hopefully also fix the darwin build.

alfreb:
Fixes #2242

- GSL span is replaced with std::span, available in C++20 / LLVM 18
- GSL Expects / Ensures is replaced with our own - the only reason for using GSL's variant was that we on a few occasions want to throw in unit tests. Not a good enough reason to keep GSL now that span is part of the standard. 
- The `<common>` header is split up into `<expects>` and `branch_prediction` renamed to `<likely>` 
- The .at() operator is replaced with [], since bounds checking is no longer done in the span. C++26 apparently fixes this.
- std::shared_ptr::unique() is removed from C++20 since it's not thread safe. It's replaced with use_count() == 1.

Note the changes in Logger - we used to inherit from the span iterator, which created a dependency on the private span implementation no longer available. I've turned this into a standalone iterator, which compiles, but is not tested - we don't yet have unit tests and there is no integration test for the logger. If the review doesn't find any obvious blunders I suggest we merge this since it passes multiple integration tests - more than we've ever tested since nix, probably. I'm working on fixing unit tests, which should give us more confidence in the logger. 

Tested with: 
- ✅ `time nix-shell --argstr unikernel ./test/net/integration/icmp6/ --run "./test.py"` 
- ✅ `time nix-shell --argstr unikernel ./test/net/integration/udp/ --run "./test.py"`
- ✅ `time nix-shell --argstr unikernel ./test/net/integration/tcp/ --run "./test.py"`
- ✅ `time nix-shell --argstr unikernel ./test/kernel/integration/smp/ --run "./test.py"`

All passing.


MagnusS:
@fwsGonzo @AndreasAakesson would you like to review or should we go ahead and merge this? I've verified that it builds and boots :-)

fwsGonzo:
LGTM

PR #2254, opened 2024-07-01
Fix memmap_test

MagnusS:
std::span does't have range checking, so the test fails. This commit just disables that part of the test.

The error was introduced when we switched from GSL::span, which had range checks.


alfreb:
Thanks!

PR #2255, opened 2024-07-03
Throw exception if Expects/Ensures fails when UNITTESTS is set

MagnusS:
Previously some unit tests would fail on calls to EXPECTS_THROW as the OS would panic. This throws an exception instead to allow those tests to proceed without failing.

### DIFF
--- a/api/expects
+++ b/api/expects
@@ -33,6 +33,7 @@
 
 #include <os.hpp>
 inline void __expect_fail(const char *expr, const char *file, int line, const char *func){
+#ifndef UNITTESTS
 #ifdef INCLUDEOS_SMP_ENABLE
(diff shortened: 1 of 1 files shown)


MagnusS:
Fixes the tests `x86_paging`, `cidr`, `test_lstack_nodes`, `test_lstack_merging` and `test_lstack_nomerge`.

MagnusS:
Force pushed an update to use `UNITTESTS` instead of `TEST` as this was more reliable and fixed a few more tests (`path_test`, `tcp_packet_test`). `UNITTESTS` is set in https://github.com/includeos/IncludeOS/blob/v0.16.0-release/test/CMakeLists.txt#L56.

PR #2256, opened 2024-07-03
Fix unittests: router_test, dhcp, os_test

MagnusS:
Fix more unit tests. See commit messages for details.

### DIFF
--- a/api/net/router.hpp
+++ b/api/net/router.hpp
@@ -168,10 +168,18 @@ namespace net {
      * @todo : Optimize!
      **/
     Route<IPV>* get_cheapest_route(typename IPV::addr dest) {
-      Routing_table all = get_all_routes(dest);
-      std::sort(all.begin(), all.end());
-      if (not all.empty()) return &all.front();
-      return nullptr;
+      Route<IPV>* match = nullptr;
+      for (auto& route : routing_table_)
+        {
+          if (route.match(dest)) {
+            if (match) {
+              match = route.cost() < match->cost() ? &route : match;
+            } else {
+              match = &route;
+            }
+          }
+        }
+      return match;
     };
 
 
--- a/test/kernel/unit/os_test.cpp
+++ b/test/kernel/unit/os_test.cpp
@@ -21,9 +21,12 @@
 
 CASE("version() returns string representation of OS version")
 {
+  // OS version should be set
   EXPECT(os::version() != nullptr);
   EXPECT(std::string(os::version()).size() > 0);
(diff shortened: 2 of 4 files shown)


api/net/router.hpp
```diff
@@ -168,10 +168,18 @@ namespace net {
      * @todo : Optimize!
      **/
     Route<IPV>* get_cheapest_route(typename IPV::addr dest) {
-      Routing_table all = get_all_routes(dest);
-      std::sort(all.begin(), all.end());
-      if (not all.empty()) return &all.front();
```
alfreb, line 173: Yea, returning a pointer to a temporary here, since get_all_routes creates a vector of fresh copies, which we then sort. Good catch. Another way of fixing it is to not return a raw pointer, but rather`std::optional<Route<IPV>>`, e.g. 

```c++
    std::optional<Route<IPV>> get_cheapest_route(typename IPV::addr dest) {
      Routing_table all = get_all_routes(dest);
      std::sort(all.begin(), all.end());
      if (not all.empty()) return all.front();
      return {};
    };
```
But your loop both avoids the copy and the issue of returning a temporary. Great!

test/kernel/unit/os_test.cpp
alfreb, line 1: Haha, because our current version is `VERY_DIRTY`, nice one 😄

test/net/unit/dhcp.cpp
```diff
@@ -19,34 +19,32 @@
 #include <net/inet>
 #include <net/interfaces>
 #include <hw/async_device.hpp>
+#include <net/dhcp/dhcpd.hpp>
 
-static std::unique_ptr<hw::Async_device<UserNet>> dev1 = nullptr;
```
alfreb, line 23: I didn't fully understand why these globals are not ok in unit tests anymore. Maybe something changed in how lest instantiates the unit tests 🤷 
Anyway, your solution also makes sure we don't have to depend on the test cases to be run in order for this to work, so it's better.

test/net/unit/router_test.cpp
alfreb, line 1: More test cases, thanks!

alfreb (approved):
Great!

test/net/unit/dhcp.cpp
MagnusS: I think it may just be that we create and store pointers to objects local to CASE() in the global static variables, and they then go out of scope (e.g. inet), but I'm not sure. It just seemed easier to put everything inside the functions to avoid that.

PR #2257, opened 2024-07-05
tests: add new targets for running valgrind and clangtidy

MagnusS:
Cleans up test/CMakeLists.txt and adds valgrind and clangtidy targets. See commit messages for details.


alfreb:
Oh, this is gold, nice! Running clangtidy pass now.

PR #2258, opened 2024-07-06
Fix wrong alignment in alloc_pmr

MagnusS:
Fixes a bug in alloc_pmr reported by valgrind. The allocator would adjust alignment to size(void*), but not size (unless it was less than size(void*)). This resulted in invalid calls to aligned_alloc(). This fixes the bug and adds a test.

I think the correct behaviour here is to always respect the alignment passed to the allocator, as C++ will use the size of the data type as alignment.

### DIFF
--- a/api/util/detail/alloc_pmr.hpp
+++ b/api/util/detail/alloc_pmr.hpp
@@ -22,6 +22,7 @@
 #include <expects>
 #include <likely>
 #include <util/units.hpp>
+#include <util/bitops.hpp>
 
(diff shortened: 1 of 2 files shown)


alfreb:
Good catch, thanks! Just to be safe I'd like to run some extensive tests before merging this, so I'd like this to come in after https://github.com/includeos/IncludeOS/pull/2260 .

alfreb:
I cherry-picked this commit on top of the testrunner now and everything is green 🍏

PR #2260, opened 2024-07-19
Add a simple test script for all working tests, fix many integration tests

alfreb:
This makes most integration tests build, and quite a few of them run.

The test script depends on vmrunner with these changes:  https://github.com/includeos/vmrunner/pull/31
See the comments there about configuring a bridge; the tests run without sudo, but requires a bridge and a qemu config file.  It should also work on hosts without kvm (TBD) as `-enable-kvm` is no longer the default for vmrunner. With all that set up, plus an internet connection, `./test.sh` should result in something like this, after ~5 min:
 
```
🚧 Step 1) Build and run unit tests
⚙️  Running this command:
nix-build unittests.nix
-------------------------------------- 💣 --------------------------------------
this derivation will be built:
  /nix/store/hwp9vsyf4ar3nbhs15zs3v32fi19xvha-unittests-dev.drv
building '/nix/store/hwp9vsyf4ar3nbhs15zs3v32fi19xvha-unittests-dev.drv'...
Running phase: unpackPhase
```
...lots of test output, and then eventually
```
<vm> ================================================================================
<vm>  IncludeOS VERY_DIRTY (x86_64 / 64-bit)
<vm>  +--> Running [ UDP test ]
<vm> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
<vm>  +--> WARNING: No good random source found: RDRAND/RDSEED instructions not available.
<vm> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
<vm>     [ Network ] Creating stack for VirtioNet on eth0 (MTU=1500)
<vm>        [ Inet ] Bringing up eth0 on CPU 0
<vm>        [ Inet ] Network configured (c0:01:0a:00:00:2a)
<vm>                 IP: 		10.0.0.55
<vm>                 Netmask: 	255.255.0.0
<vm>                 Gateway: 	10.0.0.1
<vm>                 DNS Server: 	10.0.0.1
<vm>       [ Inet6 ] Address configured fe80:0:0:0:0:0:0:4242/64
<vm> [ UDP test service ] Listening on port 4242
<Test.py> Performing UDP6 tests
<Test.py> Sent:     b'Lucky'
<Test.py> Received: b'Lucky'
<Test.py> Sent:     b'Luke'
<Test.py> Received: b'Luke'
<Test.py> Performing UDP tests
<Test.py> Sent:     b'Lucky'
<Test.py> Received: b'Lucky'
<Test.py> Sent:     b'Luke'
<Test.py> Received: b'Luke'
RECEIVED:  1472
RECEIVED:  2944
RECEIVED:  4416
RECEIVED:  5888
RECEIVED:  7360
RECEIVED:  8832
RECEIVED:  9216

[ SUCCESS ] Test completed without errors


👷💬 Step 6 succeeded 👏👏👏👏👏👏

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

🌈✨ Everything is awesome ✨
```

## Configuration:
- Use `DRY_RUN=1 ./test.sh` to show which tests will be run before running anything 
- Use `QUICK_SMOKE=1 ./test.sh` to run only a few select integration tests.


alfreb:
vmrunner is now added to the nix shell so it should no longer be needed to have that installed ahead of time. It currently points to my fork of vmrunner, until we merge this: https://github.com/includeos/vmrunner/pull/32

`./test.sh` should automagically hook this up.

test.sh
```diff
@@ -0,0 +1,94 @@
...
+    $cmd
+  done
+}
+
+# These tests should work or be removed, but are currently broken.
+exclusions=(
+  "LiveUpdate" # Missing includes
+  "context"    # Outdated - references nonexisting OS::heap_end()
+  "fiber"      # Crashes
+  "grub"       # Fails with grub-mkrescue: error: xorriso not found.
```
MagnusS, line 180: Would this succeed if we included xorriso as a dependency in the nix test build?

shell.nix
```diff
@@ -52,7 +52,7 @@ pkgs.mkShell rec {
     fi
     cmake $unikernel -DARCH=x86_64 -DINCLUDEOS_PACKAGE=${includeos} -DCMAKE_MODULE_PATH=${includeos}/cmake \
                      -DFOR_PRODUCTION=OFF
-    make -j12
+    make -j16
```
MagnusS, line 62: Perhaps you could use `$NIX_BUILD_CORES` here, see e.g. https://github.com/includeos/IncludeOS/blob/v0.16.0-release/deps/botan/default.nix#L22.

lib/LiveUpdate/CMakeLists.txt
```diff
@@ -9,25 +9,24 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
...
+    message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
+    file(DOWNLOAD "https://github.com/conan-io/cmake-conan/raw/v0.13/conan.cmake"
+      "${CMAKE_BINARY_DIR}/conan.cmake")
+  endif()
+  include(${CMAKE_BINARY_DIR}/conan.cmake)
+  conan_cmake_run(
+    CONANFILE conanfile.py
+    BASIC_SETUP
+    ${CONANPROFILE}
+  )
```
MagnusS, line 28: Do we really need this?

MagnusS (commented):
Nice work! Looks good to me, just needs to be updated to point to the right branch - other minor comments inline

shell.nix
```diff
@@ -16,7 +16,13 @@ pkgs.mkShell rec {
   includeos = pkgs.pkgsIncludeOS.includeos;
   stdenv = pkgs.pkgsIncludeOS.stdenv;
 
+  vmrunner = pkgs.callPackage (builtins.fetchGit {
+    url = "https://github.com/alfreb/vmrunner";
```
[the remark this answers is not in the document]
alfreb: Done, thanks for merging that!

test.sh
alfreb: It sure does - updated - thank you!

shell.nix
alfreb: Yep, much better - updated.

lib/LiveUpdate/CMakeLists.txt
alfreb: We sure don't :-) But what happened here is that I bulk removed the check for conanbuildinfo.cmake from everywhere to get the integration tests to build, and was focusing only on that part. Removed the remainder now.

PR #2264, opened 2024-08-22
Add back ELF symbols in stacktraces, add more tests

MagnusS:
This PR adds back stacktraces with symbol information + adds tests. See commits for additional details.

The main change is adding  `__attribute__(section(".data"))` to the static struct `relocs` used to store ELF metadata to make sure it's included in the ELF .data section (not BSS). This is needed to make sure that we don't overwrite parts of it when we initialise BSS.

On boot we first load the symbol table from `.elf_symbols` and move it to a new memory location. The new location of the symbol table is stored in `relocs.syms`: https://github.com/includeos/IncludeOS/blob/v0.16.0-release/src/kernel/elf.cpp#L350

We then initialise/clear the BSS area: https://github.com/includeos/IncludeOS/blob/v0.16.0-release/src/platform/x86_pc/kernel_start.cpp#L95

This previously had the side effect of setting `relocs.syms` back to 0 and we lost the new location of the ELF symbol table. The ELF parser initialised a few lines below would then fail.

### DIFF
--- a/cmake/os.cmake
+++ b/cmake/os.cmake
@@ -208,8 +208,10 @@ function(os_add_executable TARGET NAME)
         ${TARGET} ALL
         COMMENT "elf.syms"
         COMMAND ${ELF_SYMS} $<TARGET_FILE:${ELF_TARGET}>
+        # Copy symbols to .elf_symbols section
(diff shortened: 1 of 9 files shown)


fwsGonzo:
That's a nice find.

alfreb (approved):

alfreb:
Excellent find - just finished testing - this will be super useful!

PR #2265, opened 2024-08-24
Run tests with `nix-shell --pure`, add / fix dependencies.

alfreb:
Running test.sh on a new machine revealed some missing dependencies. Running
tests with `--pure` will ensure that only the nix provided packages can be used
during tests, improving reproducability. 

The qemu bridge setup instructions were also added to shell.nix for convenience.
`IncludeOS$ nix-shell`
should now make it clear how to configure the network bridge required for
integration tests to pass.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -76,16 +76,20 @@ pkgs.mkShell rec {
     echo $(which $CXX)
     echo -e "\nIncludeOS package:"
     echo ${includeos}
-    echo -e "\n----------------------  Qemu bridge setup  ---------------------"
+    echo -e "\n---------------------- Network privileges  ---------------------"
     echo "The vmrunner for IncludeOS tests requires bridged networking for full functionality."
-    echo "In order to use bridge networking, you need the following:"
-    echo "1. the qemu-bridge-helper needs sudo. Can be enabled with:"
+    echo "The following commands requiring sudo privileges can be used to set this up:"
+    echo "1. the qemu-bridge-helper needs sudo to create a bridge. Can be enabled with:"
     echo "   sudo chmod u+s ${pkgs.qemu}/libexec/qemu-bridge-helper"
-    echo "2. bridge43 must exist. Can be set up with \$create_bridge :"
+    echo "2. bridge43 must exist. Can be set up with vmrunner's create_bridge.sh script:"
     echo "   ${vmrunner.create_bridge}"
     echo "3. /etc/qemu/bridge.conf must contain this line:"
     echo "   allow bridge43"
-    echo "These steps require sudo. Without them we're restricted to usermode networking."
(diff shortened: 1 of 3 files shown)


shell.nix
```diff
@@ -73,5 +76,20 @@ pkgs.mkShell rec {
...
+    echo "2. bridge43 must exist. Can be set up with \$create_bridge :"
+    echo "   ${vmrunner.create_bridge}"
+    echo "3. /etc/qemu/bridge.conf must contain this line:"
+    echo "   allow bridge43"
+    echo ""
+    echo "Some tests require ping, which requires premissions to send raw packets. On some hosts"
+    echo "this is not enabled by default for iputils provided by nix. It can be enabled with:"
+    echo "4. sudo setcap cap_net_raw+ep ${pkgs.iputils}/bin/ping"
+    echo " "
+    echo
```
MagnusS, line 93: Perhaps ./test.sh should have a `--no-net` parameter. If you don't have these things set up you'll get expected failures mixed with other test failures.

test/net/integration/slaac/test.py
```diff
@@ -26,7 +26,7 @@ def Slaac_test(trigger_line):
   print(color.INFO("<Test.py>"), "Trying to ping")
   time.sleep(1)
   try:
-    command = ["ping6", "-I", "bridge43", ip_string.rstrip(), "-c",
+    command = ["ping", "-I", "bridge43", ip_string.rstrip(), "-c",
```
MagnusS, line 29:
```suggestion
    command = ["ping", "-6", "-I", "bridge43", ip_string.rstrip(), "-c",
```

To still use ipv6 you could add `-6` as a parameter to ping.

MagnusS (approved):

test/net/integration/slaac/test.py
alfreb: Yea, turns out it works for ipv6 without it - so I think this is only necessary to exclude IPv4.

shell.nix
alfreb: Yes, good point. What I would prefer even more though, is to have a smaller test suite work with userspace networking. Many of the tests will work with that, but not the IPv6 ones, and it will require some rewriting of those tests and possibly vmrunner changes, so that's a bit of work. I added a task for that here: https://github.com/includeos/IncludeOS/issues/2267

test/net/integration/slaac/test.py
MagnusS: Yes, the only benefit would be that it would fail if it accidentally was pinging an IPv4 address (which may not be likely here)

```
$ ping -6 127.0.0.1
ping: 127.0.0.1: Address family for hostname not supported
```

PR #2269, opened 2024-08-30
Memory handling improvements in early boot, fix os::panic() crash

MagnusS:
os::panic stability:
- Check that SystemLog has been initialised before writing to it. Without this check os::panic will crash if called early in the boot process.

Memory:
- Add kalloc_aligned as a replacement for memalign/aligned_alloc during early boot. These calls rely on libc and __mmap, now we use the kernel allocator directly
- Replace malloc in strdup with kalloc
- Replace memalign in IST create_stack with kalloc_aligned 
- Update mmap to give more detailed error messages and make what we support clearer

(additional details in commit messages)


alfreb:
Great finds! 

- mmap: yep, great to be more explicit about what we support. All our syscalls need a pass like that.
- Using the internal allocator instead of malloc / memalign should be mandatory. Good find.
- strdup: I didn't even know we'd implemented that ourselves😅 This change is definitely an improvement, since without it we'll explicitly call into libc before it's initialized. But I don't think we should have this in our runtime since it's a libc function provided by musl - and, it turns out - a part of C23. We should have safe C++ utilities for anything the kernel needs at different stages. But - we can do that another day!

My only objection is with the panic signature change, as this breaks the intended vmrunner functionality where we catch a panic by matching the signature, and then keep reading until we find the EOT character. This is useful for the case where a test that's supposed to work inadvertently causes a panic because of a bug. In those cases we want to inform the tester about the panic as early as possible, show the panic output, and error out without having to wait for a timeout. There's also an `on_panic` event that some disabled integration tests use, which also depends on this signature. I'll spend an hour looking into this now - since we have backtrace back we should also have a test for panic; it has been tricky to get right, some times causing double faults, so having a test for it makes sense.

MagnusS:
@alfreb happy to skip that commit, it was mainly to get it working so I could debug calls to libc during bootup with `Expects(..)`. Output stops for me right after the panic signature, so I can't see the stacktrace. Do you see the same issue with the old signature?

I added one more commit as MAP_ANONYMOUS should initialise the allocated memory to zero.

alfreb:
I had to do some work to figure out how to work on the vmrunner now that it's built with nix. I've made some changes here, making it possible to pass a local vmrunner path tho the nix shell, and revived another test category: 
 https://github.com/includeos/IncludeOS/pull/2270 

You can see how the `on_panic` event is used here: https://github.com/includeos/IncludeOS/blob/v0.16.0-release/test/stl/integration/exceptions/test.py#L32 

I'll go back to testing the original issue now, and see if I can get a backtrace to show up while still having the ability to early exit on panic and use the `on_panic` event hook in vmrunner.

MagnusS:
Great! I can no longer reproduce the problem (I'm on a different machine) - perhaps this was a local terminal issue. I'll remove it from the commit so we can merge this. If I can reproduce it later I'll take another look at it.

alfreb:
Note that a double fault does not produce a backtrace at all, by design.

MagnusS:
Updated without the panic_signature change -- `./test.sh` succeeds.

alfreb:
Thanks! For posteriority, some notes ab out the panic behaviour.  Injecting a `os::panic("Testing if panic works");` into `Service::start` in the `test/integraitostl/crt`  works, but most tests are built without functioning symbols, so we'll get this:
```
$ nix-shell ~/IncludeOS/shell.nix --pure --argstr unikernel ~/IncludeOS/test/stl/integration/crt/ --run "./test.py"
...
<vm>    [ Test CRT ] Testing C runtime
<vm> 
<vm> 
<vm> 	**** PANIC ****
<vm> CPU: 0, Reason: Testing if panic works
 <vm> Heap is at: 0x4ca000 / 0x5fa2fff  (diff=95260671)
<vm> Heap area: 1668 / 94695 Kb (allocated 1600 kb)
<vm> Total memory use: ~38% (37642240 of 100282368 b)
<vm> *** Found 0 plugin constructors:
<vm> 
<vm> *** Backtrace:symtab or strtab is empty, indicating image may be stripped
<vm> [0] 0x0000000000201353 + 0x000: 0x201353
<vm> [1] 0x000000000025be62 + 0x000: 0x25be62
<vm> [2] 0x0000000000201ae7 + 0x000: 0x201ae7
<vm> [3] 0x000000000021091d + 0x000: 0x21091d
<vm> [4] 0x00000000002ae38d + 0x000: 0x2ae38d
<vm> [5] 0x0000000000210e8a + 0x000: 0x210e8a
<vm> 

[ VM_PANIC ] CPU: 0, Reason: Testing if panic works
```

We can get a proper backtrace (thanks to @MagnusS recent work for fixing it) by setting `set(SYMBOLS ON)` in `CMakeLists.txt` in the unikernel. But then we also need to remove the `.elf.bin` postfix from the binary in `test.py`. With that we get: 

```
<vm>    [ Test CRT ] Testing C runtime
<vm> 
<vm> 
<vm> 	**** PANIC ****
<vm> CPU: 0, Reason: Testing if panic works
 <vm> Heap is at: 0x521000 / 0x5fb7fff  (diff=94990335)
<vm> Heap area: 1668 / 94431 Kb (allocated 1600 kb)
<vm> Total memory use: ~38% (37912576 of 100368384 b)
<vm> *** Found 0 plugin constructors:
<vm> 
<vm> *** Backtrace:[0] 0x00000000002010b0 + 0x2a3: os::panic(char const*)
<vm> [1] 0x000000000025be40 + 0x022: Service::start(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&)
<vm> [2] 0x00000000002017c0 + 0x327: kernel::post_start()
<vm> [3] 0x0000000000210880 + 0x09d: kernel_main
<vm> [4] 0x00000000002ae358 + 0x035: __libc_start_main
<vm> [5] 0x0000000000210e23 + 0x067: long_mode
<vm> 

[ VM_PANIC ] CPU: 0, Reason: Testing if panic works
```

Note however that `boot stl_crt` will never intercept panic at all, since intercepting panic is explicitly disabled: 
```
# Don't listen to events needed by testrunner
vm.on_success(lambda x: None, do_exit = False)
vm.on_panic(lambda x: None, do_exit = False)
```
https://github.com/includeos/vmrunner/blob/master/vmrunner/boot.py#L139C1-L141C45

I think that's what we want - boot should just give you the raw output, but when we run integration tests we'd like to exit early on unexpected panic, but also be able to write tests that explicitly triggers panic and handles that gracefully.

MagnusS:
> [..] setting `set(SYMBOLS ON)` in `CMakeLists.txt` in the unikernel. But then we also need to remove the `.elf.bin` postfix from the binary in `test.py`.

I think `SYMBOLS` here has to be `ELF_SYMBOLS`

PR #2271, opened 2024-09-02
nix: add optional ccache support

MagnusS:
Adds a new attribute `withCcache` that defaults to false. When enabled, a wrapper around clang/clang++ will be added to the IncludeOS build environment.

With an existing cache, this reduces the service build time to about 1/5 on my server when the kernel also has to be rebuilt.

The directory used for caching is /nix/var/cache/ccache and it has to exist with the right permissions. See overlay.nix for additional details.

To enable ccache, pass "--arg withCcache true" to either nix-build or nix-shell.

To manage the ccache, install "ccache" locally. Then "sudo ccache --dir /nix/var/cache/ccache --show-stats" will show cache statistics and "sudo ccache --dir /nix/var/cache/ccache --clear" will clear the cache.

Ccache is currently disabled on unittests and integration tests to make sure that these are built from a clean environment.

### DIFF
--- a/chainloader.nix
+++ b/chainloader.nix
@@ -1,6 +1,11 @@
-{ nixpkgs ? ./pinned.nix,
+{
+  withCcache ? false, # Enable ccache. Requires /nix/var/cache/ccache to exist with correct permissions.
+
+  nixpkgs ? ./pinned.nix,
   overlays ? [
-    (import ./overlay.nix)
+    (import ./overlay.nix {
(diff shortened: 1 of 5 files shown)


fwsGonzo:
What's the reason behind having to run sudo to interact with ccache here? It hasn't been required for my own projects.

MagnusS:
I've used this for a few days now and it saves a lot of time, so I'll merge this

PR #2272, opened 2024-09-03
Potential Issues and Fixes in OS Kernel Interface Functions

Bhavesh9908:
The provided code defines several functions that interface with lower-level kernel components to retrieve system state information and manage modules. However, the implementation contains potential issues, including null pointer dereferencing, unsafe type conversions, and possible concurrency issues. Specifically, the 'os::modules' function risks undefined behavior if the boot information is invalid or the module count exceeds the limits of an integer. Additionally, the lack of synchronization in multi-threaded environments could lead to race conditions. These issues can be mitigated by implementing proper checks, assertions, and synchronization mechanisms.

### DIFF
--- a/src/kernel/os.cpp
+++ b/src/kernel/os.cpp
@@ -1,10 +1,10 @@
-
 #include <os.hpp>
 #include <kernel.hpp>
 
 bool os::is_booted() noexcept {
   return kernel::is_booted();
 }
+
 const char* os::arch() noexcept {
   return Arch::name;
 }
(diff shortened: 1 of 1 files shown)


MagnusS:
The proposed changes are just minor formatting changes, so I'm closing it. If you have discovered real issues in the kernel, please report them as an issue or in a follow-up PR.

Note that the v0.16.0-release branch currently has the latest version while we are preparing a new release.

Issue #2268, opened 2024-08-28
__errno_location crash after call to __cxa_guard_acquire


MagnusS:
When enabling SMP the initial call to [`RNG::get().init()`](https://github.com/includeos/IncludeOS/blob/v0.16.0-release/src/platform/x86_pc/apic_revenant.cpp#L111) in `apic_revenant` crashes. `RNG::get().init()` returns a static variable that is initialised by the first call to `get()` ([here](https://github.com/includeos/IncludeOS/blob/v0.16.0-release/api/kernel/rng.hpp#L57)). After stepping through the code I see that the crash is caused by a surrounding call to `__cxa_guard_acquire` that protects the initialisation of the static variable, which then calls a syscall that returns `ENOSYS`. This value is written to `__errno_location` by musl, which then triggers a CPU exception.

There seems to be multiple issues here:

`__cxa_guard_acquire` is added by the compiler to protect the initialisation of static variables at runtime (such as in RNG). The implementation is [here](https://github.com/llvm/llvm-project/blob/main/libcxxabi/src/cxa_guard_impl.h). Depending on whether we have threads enabled in LIBCXX it will call various locking mechanisms in libc/musl, which have to be initialised first. If this happens early in the boot process, the kernel/musl may be unable to handle them. I'm not sure what the best approach to fix this would be, but ideally we shouldn't use code that relied on this feature until we get far enough in the boot process. 

[...]

PR #2273, opened 2024-09-04
Disallow syscalls before libc is initialised, remove all existing early boot syscalls

MagnusS:
This PR adds a panic on syscalls that happen before libc has been initialised. It also removes all the syscalls that prevented the kernel from booting after this check was added.

- During early boot stages (before initialising libc) there were several hidden calls into malloc, which then called into the brk syscall. As libc wasn't initialised this could write to unprepared areas of memory (such as errno). This is fixed by adding a new default std::pmr allocator that can be used transparently in places we need to allocate. E.g. strdup has been replaced with std::pmr::string and std::pmr::map is used in memmap.

- The kernel was compiled with threadsafe statics on, which added a guard lock around static variables that were initialised in the constructor. This again called into syscalls FUTEX and GETTID, which were not initialised. This is fixed by adding -fno-threadsafe-statics to the kernel and library builds. 

- There was a call to printf in multiboot that triggered an ioctl syscall.

See commits for additional details about each change.

Bonus:
- Adds optional elf symbols to chainloader stacktrace
- Add timeout to grub test
- Remove unused GSL dependency in chainloader.nix
- IST called memalign for SMP stacks, replaced with kalloc_align


fwsGonzo:
I think we relied heavily upon brk during startup before, which was probably fine. The SMP memalign call was unfortunate.

MagnusS:
> I think we relied heavily upon brk during startup before, which was probably fine. The SMP memalign call was unfortunate.

It may have worked fine (and maybe it still does), but it seems much safer to avoid all the early syscalls if we can. There's also the write to errno in syscall_ret for instance, which we know may cause crashes (from #2268). 

The brk calls come from [`__expand_head`](https://git.musl-libc.org/cgit/musl/tree/src/malloc/expand_heap.c?h=v1.1.18&id=eb03bde2f24582874cb72b56c7811bf51da0c817) in musl `malloc` -- the kernel has already called  _init_brk() when it initialised the heap, but init_libc isn't called yet. So it really depends on what happens on the musl side.

PR #2274, opened 2024-09-06
Restore basic SMP support

MagnusS:
This restores basic SMP-support. The main change is that the kernel now uses explicit spinlocks for static variables initialised with get()-methods. This is necessary after we disabled threadsafe statics (which may not have worked properly in the past, as libc wasn't initialised).

I've updated the default PMR allocator to be SMP safe and replaced most of the dynamic datastructures in SMP/APIC with PMR-versions. This avoids calls through libc, although the allocator itself could probably be made more efficient.

The SMP test now only tests for tasks running on different cores and does a malloc/stack/PMR allocation test. We don't support printf or timers from different cores yet.

Without the locks the initialisation of the SMP cores would occasionally crash and reboot or the test itself would deadlock. I've run this 40-50 times with 16 cores now without a crash, but it would be very useful with a read-through by someone familiar with the code and AP init to double check that it makes sense!

Bonus:
- Cherrypicked nanosleep syscall implementation by fwsGonzo from f9828aed29e57e72c0eb9b8c785152c8b7135d3d

Some other changes that may be useful to bring in later:
- https://github.com/includeos/IncludeOS/commit/47f6702aefb7f007f3f2764dff7becc01e036dfd
- https://github.com/includeos/IncludeOS/commit/6944de0ececd7e842d7a050afcb8300765750a84
- https://github.com/includeos/IncludeOS/commit/330432b39d41b9804437f3fc18182c17b24d57ba

### DIFF
--- a/api/kernel/rng.hpp
+++ b/api/kernel/rng.hpp
@@ -23,6 +23,7 @@
 #include <cstdlib>
 #include <cstdint>
 #include <delegate>
+#include <smp_utils>
 
 // Incorporate seed data into the system RNG state
 extern void rng_absorb(const void* input, size_t bytes);
@@ -54,14 +55,18 @@ class RNG : public FD_compatible {
 public:
   static RNG& get()
   {
-    static RNG rng;
+#ifdef INCLUDEOS_SMP_ENABLE
+    static Spinlock lock_;
+    std::lock_guard<Spinlock> lock(lock_);
+#endif
+    static RNG rng = RNG();
     return rng;
   }
 
   static void init();
 
 private:
-  RNG() {}
+  RNG() {};
 
 };
 
--- a/api/smp
+++ b/api/smp
@@ -52,7 +52,7 @@ public:
   static int cpu_count() noexcept;
 
   // Return the indices of all initialized CPU cores
-  static const std::vector<int>& active_cpus();
+  static const std::pmr::vector<int>& active_cpus();
 
   static int active_cpus(int index){
     return active_cpus().at(index);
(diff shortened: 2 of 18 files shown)


fwsGonzo:
Looks very nice. Great work. The timer system is supposed to be per-cpu, so locking there doesn't make sense. Is it required for a test to pass? The timer system internally uses `get()` or something, which should access a per-cpu struct.

MagnusS:
It looks like you're right! It seems to still pass the test reliably without the timer locks. I added those locks because I was trying to get the timer code to work as well, but there's more work needed for that -- do you think Timers::* can be called without locks from the SMP::add_task handler as long as get() is synchronized? If they can I'll just remove that last commit

MagnusS:
There seems to still be a race when the APs are initialised, but I'm unable to reproduce it reliably - but it does occasionally happen. It also happens with the Timer locks, so I think it's unrelated.

I noticed that you added a lock and moved sti later in this old commit @fwsGonzo https://github.com/includeos/IncludeOS/commit/284c92cfb83381e2991460c6b40edea245b9b1a9 -- so perhaps this is an older issue that we have to solve separately to this PR. I've not seen any crashes if the kernel gets through AP init.

MagnusS:
I removed the locks from Timers:: and added a static spinlock to protect part of the AP init. I then ran the test 100 times with this command: `for f in $(seq 1 100); do echo $f; date ; nix-shell --pure --argstr unikernel ./test/kernel/integration/smp --arg withCcache true --run ./test.py ; done`. There were no failures. 

I can't guarantee that it doesn't crash after 101 tests of course, but I think this should be fine to merge - then we can continue to work on these issues if they reappear

PR #2275, opened 2024-09-07
test: optionally enable ccache for test runs

alfreb:


MagnusS:
Nice to have a test flag for this! But I think to avoid a "broken default" for some users it should be disabled by default. Ccache requires local nix configuration changes before it will work. Then it would also match how it's used for nix-build/nix-shell.

There's also at least a theoretical chance that ccache is too aggressive and gives you old build artefacts, so could be useful to encourage running the final tests without it unless you know you really want to use it :-)

alfreb:
Sure, we can do that.

PR #2278, opened 2024-09-08
Add SMP option to nix, enable SMP for selected tests, add ccache option for tests.

alfreb:
- test.sh: Add env var to enable ccache, default to off.
- Add smp parameter to nix default, shell, and overlay. Enable with --arg smp true.
- Added multiple cores to paging, UDP and exceptions tests, add them to an SMP test set.
- Enable boot logger to tests make it easier toverify that multiple cores come up.
- Disable SMP for chainloader. It would be nice to test 32 bit multicore, but not in there.


MagnusS:
Looks great - thanks for the rebase :)

PR #2280, opened 2024-09-13
Test improvements

MagnusS:
Contains mostly minor test cleanup. 

- Updates the grub test to build the iso during the build-phase instead of during the test run. This makes dependency management easier and doesn't require write access during test run, which is useful if we want to run it in a sandbox or VM.

- `kprint` is moved from `src/kernel/include` to `api` to allow it to be tested in the integration tests without having access to the original source. A side-effect is that kprintf becomes available for all services (but they have to `#include <kprint>`).

Other
- Fix shellcheck reported quoting errors in grubiso.sh (as it contains an rm -rf)
- Remove old conan files / scripts from test/integration
- Remove unused GSL dependency from example
- Add ccache support when building unittests

(Tested with `./test.sh`)


alfreb:
Nice cleanup, thanks!

PR #2281, opened 2024-09-13
Add support for cmake find_package

MagnusS:
Enables `find_package()` from other nix-installed packages by adding `cmake/includeos-config.cmake.in` template, which is then installed to `cmake/includeos-config.cmake` with the correct paths filled in. This avoids the cmake flags that previously had to be set manually. Updates `example.nix` to use the new method.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.6.0)
+cmake_minimum_required(VERSION 3.6)
 
 #we are only creating libraries for ELF
 set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
@@ -16,6 +16,14 @@ add_subdirectory(src)
 install(FILES cmake/linux.service.cmake DESTINATION cmake)
 install(FILES cmake/os.cmake DESTINATION cmake)
 install(FILES cmake/includeos.cmake DESTINATION cmake)
+
+# Install config to enable find_package(IncludeOS)
+include(CMakePackageConfigHelpers)
+configure_package_config_file(
+  cmake/includeos-config.cmake.in
+  ${CMAKE_CURRENT_BINARY_DIR}/cmake/includeos-config.cmake
+  INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/cmake/includeos-config.cmake DESTINATION cmake)
 
 # Install IncludeOS user code dependencies
 install(FILES src/service_name.cpp DESTINATION src)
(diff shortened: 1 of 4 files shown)


alfreb:
Right, so this produces the following: 

```
$ cat $(nix-build ~/IncludeOS)/cmake/includeos-config.cmake 
# Config file to be picked up by find_package()

####### Expanded from @PACKAGE_INIT@ by configure_package_config_file() #######
####### Any changes to this file will be overwritten by the next CMake run ####
####### The input file was includeos-config.cmake.in                            ########

get_filename_component(PACKAGE_PREFIX_DIR "${CMAKE_CURRENT_LIST_DIR}/../../" ABSOLUTE)

macro(set_and_check _var _file)
  set(${_var} "${_file}")
  if(NOT EXISTS "${_file}")
    message(FATAL_ERROR "File or directory ${_file} referenced by variable ${_var} does not exist !")
  endif()
endmacro()

macro(check_required_components _NAME)
  foreach(comp ${${_NAME}_FIND_COMPONENTS})
    if(NOT ${_NAME}_${comp}_FOUND)
      if(${_NAME}_FIND_REQUIRED_${comp})
        set(${_NAME}_FOUND FALSE)
      endif()
    endif()
  endforeach()
endmacro()

####################################################################################
set(ARCH "x86_64")
set(INCLUDEOS_PACKAGE /nix/store/3r5nc4b2rirhzqmhmjx0mwjc4v0f3wj2-includeos-static-x86_64-unknown-linux-musl-dev)
set(CMAKE_MODULE_PATH /nix/store/3r5nc4b2rirhzqmhmjx0mwjc4v0f3wj2-includeos-static-x86_64-unknown-linux-musl-dev/cmake)
```
So `3r5nc4b2rirhzqmhmjx0mwjc4v0f3wj2` is the hash for my IncludeOS package and if I add IncludeOS as a nix dependency it will hook in the path for that to `find_package` via `CMAKE_PREFIX_PATH`. We can verify this by going into the nix shell for the example and inspecting its environment: 
```
$ nix-shell example.nix
$ [nix-shell:~/IncludeOS]$ env | grep 3r5nc4b2rirhzqmhmjx0mwjc4v0f3wj2
CMAKE_INCLUDE_PATH=... :/nix/store/q3g3nixd69jvrwcz9qd60xmm5f8wq1ig-gtest-static-x86_64-unknown-linux-musl-1.14.0:/nix/store/3r5nc4b2rirhzqmhmjx0mwjc4v0f3wj2-includeos-static-x86_64-unknown-linux-musl-dev...
...
CMAKE_PREFIX_PATH=/nix/store/ag0fwkx5sirdlyx0maz55i636rh5lhxm-libcxx-static-x86_64-unknown-linux-musl-18.1.6-dev: ... :/nix/store/3r5nc4b2rirhzqmhmjx0mwjc4v0f3wj2-includeos-static-x86_64-unknown-linux-musl-dev
```

So we should be able to make a locally built package and instead of exporting `INCLUDEOS_PACKAGE` pointing to it, we  to add it to `CMAKE_PREFIX_PATH` or install it to a location already known to cmake. 

Nice!

PR #2282, opened 2024-09-14
Remove unused files after switching to nix

MagnusS:
- Remove more conan references, all conanfile.py scripts.
- Removes various shell scripts that we no longer use
- Removes the vagrant setup, Jenkinsfile and Dockerfile
- Removes unused source code and linker script.
- Removes unused dependencies from deps.

It may be useful to recreate some of this later (e.g. docker, vagrant), but it will have to substantially rewritten to use nix instead of conan - so I think it's better to just remove it for now.


alfreb:
Thanks! I think there's still a path to building with conan and I'm not against that, in which case the deps/.../conanfile's are probably a reasonable starting point as they're not that old. But I agree that we shouldn't keep it around when it's not maintained. If someone wants to get it working again they can restore them from git history. @KristianJerpetjon FYI.

PR #2283, opened 2024-09-15
Revive Fieldmedic plugin, add early kernel diagnostic hooks 

alfreb:
The intention is to create a reusable pattern for kernel self tests and diagnostics that costs nothing default. We have quite a few defensive validation steps in the kernel added organically as issues have been found. Ideally a fast booting kernel should always do the correct thing without having to validate. In practice, validation is crucial. 

With this proposed setup we can optionally add a validation plugin (fieldmedic) that performs more extensive self tests after key checkpoints in the unikernel lifetime has been reached, with negligible cost in the default case (the cost of an empty function call per hook) and with zero cost when toggled off at compile time.

See commit messages for details. 

To try it out: 
```
$ nix-shell ~/IncludeOS/shell.nix --pure  --argstr unikernel ~/IncludeOS/test/kernel/integration/osinit --run ./test.py
...
<vm> <Multiboot>OS loaded with 1 modules
<vm> 	* osinit.elf.bin booted with vmrunner @ 0x29e000 - 0x43f6c0, size: 1709760b
<vm> * Multiboot begin: 0x9500
<vm> * Multiboot end: 0x3e5490
<vm> [ Field Medic ] ⛑️  BSS Diagnostic passed
<vm> [x86_64 PC] constructor
<vm> [ Machine ] Initializing heap
<vm> [ Machine ] Main memory detected as 129998656 b
<vm> [ Machine ] Reserving 1048576 b for machine use
<vm> [ Field Medic ] ⛑️  Machine allocator for x86 PC functional
<vm> [ Field Medic ] ⛑️  Elf header intact, global constructors functional
<vm> [ Field Medic ] ⛑️  Malloc and brk are functional
<vm> ================================================================================
<vm> 
<vm>                            #include<os> // Literally
<vm> 
<vm> ================================================================================
<vm>      [ Kernel ] Stack: 0x1ffbb8
<vm>      [ Kernel ] Boot magic: 0x2badb002, addr: 0x9500
...
...
<vm>      [ Kernel ] Running service constructors
<vm> --------------------------------------------------------------------------------
<vm> ================================================================================
<vm>  IncludeOS VERY_DIRTY (x86_64 / 64-bit)
<vm>  +--> Running [ OS initialization test ]
<vm> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
<vm>  +--> WARNING: No good random source found: RDRAND/RDSEED instructions not available.
<vm>  +-->        To make this warning fatal, re-compile with FOR_PRODUCTION=ON.
<vm> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
<vm> Service::start entered
<vm> [ Field Medic ] ⛑️  Service finished. Diagnosing.
<vm>                 [+] Field medic plugin active
<vm>                 [+] Post .bss invariant still holds
<vm>                 [+] Post machine init invariant still holds
<vm>                 [+] Post init libc invariant still holds
<vm> [ Field Medic ] ⛑️  Diagnose complete. Healthy ✅

[ SUCCESS ] OS initialization test succeeded
```

### DIFF
--- a/api/info
+++ b/api/info
@@ -29,7 +29,7 @@
 #define INFO2(TEXT, ...) kprintf("%16s" TEXT "\n"," ", ##__VA_ARGS__)
 #define CENTER(TEXT) kprintf("%*s%*s\n",LINEWIDTH/2 + (int)strlen(TEXT)/2,TEXT,LINEWIDTH/2-(int)strlen(TEXT)/2,"")
 #include <string>
-#define FILLINE(CHAR) kprintf("%s\n",std::string(LINEWIDTH,CHAR).c_str())
+#define FILLINE(CHAR) kprintf("%s\n",std::pmr::string(LINEWIDTH,CHAR).c_str())
 #define CAPTION(TEXT) FILLINE('='); kprintf("\n"); CENTER(TEXT); kprintf("\n"); FILLINE('=')
 
 // Checkboxes - for initialization output and testing
--- a/api/info
+++ b/api/info
@@ -21,19 +21,21 @@
 
 #define LINEWIDTH 80
 
+#include <kprint>
+
 // Informational prints
 #ifndef NO_INFO
-#define INFO(FROM, TEXT, ...) printf("%13s ] " TEXT "\n", "[ " FROM, ##__VA_ARGS__)
-#define INFO2(TEXT, ...) printf("%16s" TEXT "\n"," ", ##__VA_ARGS__)
-#define CENTER(TEXT) printf("%*s%*s\n",LINEWIDTH/2 + (int)strlen(TEXT)/2,TEXT,LINEWIDTH/2-(int)strlen(TEXT)/2,"")
+#define INFO(FROM, TEXT, ...) kprintf("%13s ] " TEXT "\n", "[ " FROM, ##__VA_ARGS__)
+#define INFO2(TEXT, ...) kprintf("%16s" TEXT "\n"," ", ##__VA_ARGS__)
+#define CENTER(TEXT) kprintf("%*s%*s\n",LINEWIDTH/2 + (int)strlen(TEXT)/2,TEXT,LINEWIDTH/2-(int)strlen(TEXT)/2,"")
 #include <string>
-#define FILLINE(CHAR) printf("%s\n",std::string(LINEWIDTH,CHAR).c_str())
-#define CAPTION(TEXT) FILLINE('='); printf("\n"); CENTER(TEXT); printf("\n"); FILLINE('=')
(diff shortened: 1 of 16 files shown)


alfreb:
Oups, good catch, thank you! Fixed.

api/info
```diff
@@ -21,19 +21,21 @@
...
-#define INFO(FROM, TEXT, ...) printf("%13s ] " TEXT "\n", "[ " FROM, ##__VA_ARGS__)
-#define INFO2(TEXT, ...) printf("%16s" TEXT "\n"," ", ##__VA_ARGS__)
-#define CENTER(TEXT) printf("%*s%*s\n",LINEWIDTH/2 + (int)strlen(TEXT)/2,TEXT,LINEWIDTH/2-(int)strlen(TEXT)/2,"")
+#define INFO(FROM, TEXT, ...) kprintf("%13s ] " TEXT "\n", "[ " FROM, ##__VA_ARGS__)
+#define INFO2(TEXT, ...) kprintf("%16s" TEXT "\n"," ", ##__VA_ARGS__)
+#define CENTER(TEXT) kprintf("%*s%*s\n",LINEWIDTH/2 + (int)strlen(TEXT)/2,TEXT,LINEWIDTH/2-(int)strlen(TEXT)/2,"")
 #include <string>
-#define FILLINE(CHAR) printf("%s\n",std::string(LINEWIDTH,CHAR).c_str())
-#define CAPTION(TEXT) FILLINE('='); printf("\n"); CENTER(TEXT); printf("\n"); FILLINE('=')
+#define FILLINE(CHAR) kprintf("%s\n",std::string(LINEWIDTH,CHAR).c_str())
```
MagnusS, line 32: std::string may use malloc and requires heap to be initialised, so you could use`std::pmr::string` instead (which should be set up by that point)

MagnusS:
Everything compiles and tests are green. I noticed a call to std::string that I missed on first review - I'm not sure how early FILLLINE is called, but you may want to replace that with `std::pmr::string` to avoid any calls to malloc before libc is initialised.

api/info
alfreb: > std::string may use malloc and requires heap to be initialised, so you could use`std::pmr::string` instead (which should be set up by that point)

Good point! And funny that this is used exactly at the point when std::string is possible:  

```
<vm> [ Field Medic ] ⛑️  Malloc and brk are functional
<vm> ================================================================================
```
But much better to make it safe, will do.

alfreb: Done!

PR #2300, opened 2025-09-18
Update to nixpkgs 25.05

MagnusS:
- Move pin to nixpkgs 25.05
- Botan 2 now builds without overriding the configurePhase (and fails with the previous override)
- Depends on https://github.com/includeos/vmrunner/pull/39

./test.sh tests succeed, except network tests (due to missing tap device on this host)


alfreb (approved):
Nice!

PR #2306, opened 2025-10-07
fix deprecation warning: preceding whitespace to operator

mazunki:
https://cplusplus.github.io/CWG/issues/2521.html


mazunki:
While going through the deprecation warning, i noticed there was an inconsistent use of spaces vs not spaces between the operator symbol throughout the code. It seems most style guides recommend omitting the space, so i went with that.

Issue #2310, opened 2025-10-10
Impure path used during linking of unikernel image


mazunki:
Attempting to build `example/` (through `nix-shell ./shell.nix`) on Gentoo gives the following errors:

```log
ld.lld: error: /nix/store/wkn6ivpdjri06sjig43k43n72cidzb4p-includeos-static-x86_64-unknown-linux-musl-dev/libcxx/lib/libc++.a(cxa_exception_storage.cpp.o): unable to find library from dependent library specifier: pthread
ld.lld: error: /nix/store/wkn6ivpdjri06sjig43k43n72cidzb4p-includeos-static-x86_64-unknown-linux-musl-dev/libcxx/lib/libc++.a(fallback_malloc.cpp.o): unable to find library from dependent library specifier: pthread
ld.lld: error: /nix/store/wkn6ivpdjri06sjig43k43n72cidzb4p-includeos-static-x86_64-unknown-linux-musl-dev/libcxx/lib/libc++.a(memory.cpp.o): unable to find library from dependent library specifier: pthread
ld.lld: error: /nix/store/wkn6ivpdjri06sjig43k43n72cidzb4p-includeos-static-x86_64-unknown-linux-musl-dev/libcxx/lib/libc++.a(mutex.cpp.o): unable to find library from dependent library specifier: pthread
ld.lld: error: /nix/store/wkn6ivpdjri06sjig43k43n72cidzb4p-includeos-static-x86_64-unknown-linux-musl-dev/libcxx/lib/libc++.a(cxa_guard.cpp.o): unable to find library from dependent library specifier: pthread
ld.lld: error: /nix/store/wkn6ivpdjri06sjig43k43n72cidzb4p-includeos-static-x86_64-unknown-linux-musl-dev/libcxx/lib/libc++.a(condition_variable.cpp.o): unable to find library from dependent library specifier: pthread
ld.lld: error: /nix/store/wkn6ivpdjri06sjig43k43n72cidzb4p-includeos-static-x86_64-unknown-linux-musl-dev/libcxx/lib/libc++.a(thread.cpp.o): unable to find library from dependent library specifier: pthread
ld.lld: error: /nix/store/wkn6ivpdjri06sjig43k43n72cidzb4p-includeos-static-x86_64-unknown-linux-musl-dev/libunwind/lib/libunwind.a(libunwind.cpp.o): unable to find library from dependent library specifier: dl
[...]

PR #2311, opened 2025-10-13
Update deprecated cmake version

mazunki:
As discussed in #2310, there are some issues with the build system picking a linker for unikernels, even in `--pure` mode.

Passing `set(CMAKE_LINKER_TYPE LLD)`  seems to fix this, but doesn't work in older versions of CMake. This PR updates the minimum version of cmake we depend on to 3.31.6 (as reported by `nix-shell --pure ./develop.nix --run 'cmake --version'`).

Bumping this version reports two usages of `PRE_BUILD` which is invalid in modern CMake (I believe they always have been, but just been silently ignored? https://cmake.org/cmake/help/latest/command/add_custom_command.html), and only be relevant to the Visual Studio editor and build system. I've simply removed them (commit 7097eda).

Lastly, I have added an explicit dependency to LLD/LLVM (https://cmake.org/cmake/help/latest/variable/CMAKE_LINKER_TYPE.html) in the example unikernel. With CMake updated, the unikernel links successfully even on Gentoo .

### DIFF
--- a/lib/LiveUpdate/CMakeLists.txt
+++ b/lib/LiveUpdate/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.6.0)
+cmake_minimum_required(VERSION 3.31.6)
 
 set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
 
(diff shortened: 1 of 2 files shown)


alfreb:
This is fine. I generally don't want to deprecate tools that work just fine, and are supported. But the only build system we support is nix with the pinned versions, and I verified now that this is 3.31.6, so let's go with that, even though it's not technically the minimum version.

PR #2315, opened 2025-10-14
suppress verbose --target mismatch warning

mazunki:
`nixpkgs/pkgs/build-support/cc-wrapper/add-clang-cc-cflags-before.sh` has a flag to suppress the rather annoying (albeit correct) warning:

```
Warning: supplying the --target x86_64-pc-linux-elf != x86_64-unknown-linux-musl argument to a nix-wrapped compiler may not work correctly - cc-wrapper is currently not designed with multi-target compilers in mind. You may want to use an un-wrapped compiler instead.
```

This adds an option (enabled by default across the board) to suppress this. Because of how `example/` is built, I had to implement it two places (ideally I'd inherit it in `example.nix`, which is only used as a test currently... maybe we should change that).

### DIFF
--- a/chainloader.nix
+++ b/chainloader.nix
@@ -6,6 +6,7 @@
     (import ./overlay.nix {
       inherit withCcache;
       smp = false; # No SMP for chainloader
+      disableTargetWarning = true;
     })
   ],
   pkgs ? import nixpkgs {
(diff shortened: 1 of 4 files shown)


alfreb:
Interesting. Do we know where  `--target x86_64-pc-linux-elf != x86_64-unknown-linux-musl` comes from? I see we're setting `${ARCH}-pc-linux-elf` multiple places - should that be `x86_64-unknown-linux-musl` instead? See e.g. :

```
cmake/includeos.cmake:set(TRIPLE "${ARCH}-pc-linux-elf")
cmake/library.cmake:set(TRIPLE "${ARCH}-pc-linux-elf")
```

mazunki:
Fixed and tested. @MagnusS

mazunki:
Agreed. I can open one.

PR #2316, opened 2025-10-17
Refactor how we build unikernels for testing

### LINKED ISSUE #2310: Impure path used during linking of unikernel image

mazunki:
In order to fix #2310, I have migrated the building of the unikernels to nix instead of manually calling cmake in a shellHook.

This changes how we run tests. nix has a `checkPhase` specifically designed to test builds. We now use this whenever we can. Because we're now using `nix-build` for all the building, this implicitly becomes pure.

Another change with this is that we will no longer need to rebuild and recheck tests we already know are passing. We could enable `--check` to force rechecking unchanged derivations. Do we want that?
Furthermore, we should now be able to pass an architecture target for testing.

Annoyingly, not all tests are able to run through the checking phase because capabilities (specifically, we need `cap_net_admin` and `cap_net_raw` on `qemu-bridge-helper`) are dropped upon entering this phase. To work around this, we run the few tests that actually need to be run without a proper reproducible sandbox in a shellHook (this is partially what we were doing before). Ideally, we'd get rid of this. I have seen that [security.wrappers.\<name\>.capabilities](https://search.nixos.org/options?channel=25.05&query=Security.wrappers) exists: maybe we can handle the capabilities through this instead: but I suppose that is part of [includeos/vmrunner](https://github.com/includeos/vmrunner) instead.

----
I have also renamed `example.nix` to `unikernel.nix` because it's not really specific to `example/` at all. I considered moving it to `tests/default.nix` (it would make `./test.sh` slightly cleaner), instead, but wasn't sure how we feel about "hiding" these entry points. Personally I have a preference towards keeping the root directory uncluttered.

`shell.nix` is not involved in testing at all anymore, but it does provide some useful PATHs upon entering the shell. Maybe we can merge `develop.nix` back into `shell.nix` after this...

In summary:
- Fixes and closes #2310
- Prefer `checkPhase` over `shellHook` when possible
- `shell.nix` is now mostly a stub
- Rename `example.nix` to `unikernel.nix`
- Handle building of test services in `unikernel.nix` through nix artifacts
- Update the motd/README upon entering a shell (`cap_net_admin+ep qemu-bridge-helper` was missing!)

All tests are passing.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -80,8 +80,8 @@ multicore_subset(){
 }
 
 smoke_tests(){
-  nix-build ./unikernel.nix $CCACHE_FLAG --argstr unikernel ./test/net/integration/udp
-  nix-build ./unikernel.nix $CCACHE_FLAG --argstr unikernel ./test/net/integration/tcp
+  nix-build ./unikernel.nix $CCACHE_FLAG --argstr unikernel ./test/net/integration/udp --arg doCheck true
+  nix-build ./unikernel.nix $CCACHE_FLAG --argstr unikernel ./test/net/integration/tcp --arg doCheck true
   nix-build ./unikernel.nix $CCACHE_FLAG --argstr unikernel ./test/kernel/integration/paging --arg doCheck true
   nix-build ./unikernel.nix $CCACHE_FLAG --argstr unikernel ./test/kernel/integration/smp --arg doCheck true
 }


example.nix
alfreb: Yep, unikernel.nix makes more sense for what it's becoming now.

shell.nix
```diff
@@ -48,49 +44,44 @@ includeos.pkgs.mkShell.override { inherit (includeos) stdenv; } rec {
   ];
 
   shellHook = ''
+    cat <<-EOF
```
alfreb: Yep, cleaner, thanks.

test.sh
```diff
@@ -223,7 +238,9 @@ exclusions=(
   "websocket" # Linking fails, undefined ref to http_parser_parse_url, http_parser_execute
 )
 
+sandboxed=false
```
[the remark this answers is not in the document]
mazunki: > I thought they didn't require capabilities themselves to run, 

vmrunner requires permission to attach to the bridge. I'm pretty sure it's vmrunner's startup that requires the capability, not the unikernel or messaging.

mazunki:
I think I've addressed the issues mentioned. Did I miss something?

alfreb:
> I think I've addressed the issues mentioned (tests are still passing, hopefully this time without skipping any). Did I miss something?

All good from my side, but I requested a review from @MagnusS as well as mentioned

MagnusS (changes requested):
Great, I think this is an improvement over the previous solution - it's much better to run these in `nix-build`. Changes look good to me, but see minor comments inline. It also looks like the section about "Running tests" in README.md will need a small update. Regarding your question about `--check` -- as long as tests rerun when the kernel has been changed I think it's fine to not add this (we can add it later if needed)

PR #2317, opened 2025-10-18
Bump c++17 to c++20 on linux.service.cmake

mazunki:
This is a rather simple PR that updates linux.service.cmake to use c++17. With this, all `-std=` and `CXX_STANDARD` references are using c++20. 

I wonder why this was never updated to c++20. I would like to clean up and document the `cmake/*.cmake` files. As an outsider, it's very difficult to understand what each of the files are responsible for, and how they're tied together (e.g. transitive dependencies). I will work on this in its own PR at a later stage.

I noticed there was a stary `compile_options.txt` file here. I've deleted it since it is misleading. I was not able to find the commit in which the reference to it was deleted: I could only find when it was added. Is the git history corrupted, or am I failing to use git correctly? `git log -G 'compiler_options\.txt' --` only shows one entry from 2016-10-29 (342aad79e), and nothing more recent.

Separately, we should probably also refactor a few things related to CMake: currently we add `-std=` manually (not portable!) and other times we define `CMAKE_CXX_VERSION`. Furthermore, it's a bit weird (maybe not wrong) that there's so much boilerplate for building C files:

```shell
$ find -name '*.c'
./src/crt/c_abi.c
./src/crt/ctype_b_loc.c
./src/crt/ctype_tolower_loc.c
./src/crt/pthread.c
./src/util/memstream.c
./test/posix/integration/conf/test_pathconf.c
./test/posix/integration/conf/test_pwd.c
./test/posix/integration/conf/test_sysconf.c
$ find -name '*.c' | wc -l
8
```

- Should those be ported to C++? Do we want to support both C and C++? I can see the argument both ways.
- Is the `pthread.c` meant to exist in this branch?

### DIFF
--- a/cmake/linux.service.cmake
+++ b/cmake/linux.service.cmake
@@ -2,9 +2,9 @@
 #   Linux Userspace CMake script   #
 ####################################
 
-#set(CMAKE_CXX_STANDARD 17)
+#set(CMAKE_CXX_STANDARD 20)
 set(COMMON "-g -O2  -Wall -Wextra")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 ${COMMON}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++20 ${COMMON}")
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${COMMON}")
 
 option(BUILD_PLUGINS "Build all plugins as libraries" OFF)


mazunki:
All tests are passing with this.

alfreb:
Good find! No, I don't think there is a legit reason to have both C and C++. I think the reason it was there is becuase, well, crt means the C runtime, and bootstrapping that is much much simpler than bootstrapping C++ (especially getting exceptions to work was hard). But now that we have a pretty stable base I think it makes sense to convert all this to C++. 

I created this issue for it now
https://github.com/includeos/IncludeOS/issues/2323

PR #2324, opened 2025-10-19
Silence unused warnings related to tests

mazunki:
Continuation of #2320 , but addressing tests.

@alfreb `test/kernel/integration/paging/service.cpp` has an unused `heap_code` variable. I see you updated this test in 2019.

The `void(*heap_code)() = (void(*)()) malloc(42);` line is at the very beginning of the test. I suspect the intention of this is to warm of the allocator. Did you mean to change this to `magic->heap_code`?

I don't want to touch the tests based on assumptions on what the intention was, without fully understanding what the idea behind it was.

Additionally, `Expects(magic->last_error = Page_fault);` is surely a subtle bug? That `=` should be `==`, right?

### DIFF
--- a/api/net/inet.hpp
+++ b/api/net/inet.hpp
@@ -535,4 +535,34 @@ namespace net {
   };
 }
 
+namespace net::ip4 {
+  inline bool operator==(const net::ip4::Addr& lhs, const net::Inet& rhs) noexcept {
+    return lhs == rhs.ip_addr();
+  }
+  inline bool operator==(const net::Inet& lhs, const net::ip4::Addr& rhs) noexcept {
+    return lhs.ip_addr() == rhs;
+  }
+  inline bool operator!=(const net::ip4::Addr& lhs, const net::Inet& rhs) noexcept {
+    return !(lhs == rhs);
+  }
+  inline bool operator!=(const net::Inet& lhs, const net::ip4::Addr& rhs) noexcept {
+    return !(lhs == rhs);
+  }
+}  // namespace net::ip4
+
+namespace net::ip6 {
+  inline bool operator==(const net::ip6::Addr& lhs, const net::Inet& rhs) noexcept {
+      return lhs == rhs.ip6_addr();
+  }
+  inline bool operator==(const net::Inet& lhs, const net::ip6::Addr& rhs) noexcept {
+      return lhs.ip6_addr() == rhs;
+  }
+  inline bool operator!=(const net::ip6::Addr& lhs, const net::Inet& rhs) noexcept {
+      return !(lhs == rhs);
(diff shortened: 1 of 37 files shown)


mazunki:
This has been rebased and is ready for review. As the title mentions, these tests are related to tests; but touches upon some source too.

There are still a few warnings I'd like to address, but they're more related to issues with the source itself, so I will handle them at a later point

alfreb:
> The void(*heap_code)() = (void(*)()) malloc(42); line is at the very beginning of the test. I suspect the intention of this is to warm of the allocator. Did you mean to change this to magic->heap_code?

The point is just to get a pointer to an area owned by malloc, which we know will be both readable and writable. And then we want to try to execute code in that area, which should fail when we've execute proected the same area.

The unused function looks like it can be deleted - I might have started developing with that, then moved it into a struct without deleting it. So feel free to delete to get rid of dead code.

api/net/inet.hpp
```diff
@@ -535,4 +535,34 @@ namespace net {
   };
 }
 
+namespace net::ip4 {
+  inline bool operator==(const net::ip4::Addr& lhs, const net::Inet& rhs) noexcept {
```
alfreb: I don't understand these - why are we comparing an address to an IP stack?

mazunki:
It's been a while since I touched these, but seems like they're still working. :)

test/unit/fs/unit_fat.cpp
```diff
@@ -40,6 +40,11 @@ CASE("Initialize FAT fs")
     [&lest_env] (auto err, File_system& fs)
     {
       EXPECT(!err);
+      EXPECT(fs.is_valid() == true);
+
+      // EXPECT(fs.name() == "FAT32"); // FIXME: test fails
```
MagnusS, line 45: Did this work before your changes? Why is it failing now?

MagnusS (commented):
This generally looks good to me, but I haven't run the network tests yet. I also left a comment regarding the FAT32 check. Have you verified that network tests are ok?

test/unit/fs/unit_fat.cpp
mazunki: It was never tested. It only tested that there was no errors. As I went through these warnings I verified if stuff was missing from the tests, which was the case here.

mazunki: The file loaded here is `memdisk.fat`, which has the bytes `FAT32` at bytes 56-60.

mazunki:
I always test with network tests enabled, but I can double check in a couple hours :)

mazunki:
Yep, all good.

PR #2325, opened 2025-10-19
Dependency path wrapping + cleanup

mazunki:
In order to more easily get clangd happy, I've set up `${dependency}.include` and `${dependency}.lib` targets to all `deps/` (except lest, which is header-only. We should consider precompiling it), instead of requiring to guess these paths every time they're used.

I've also nixified `deps/libfdt`. Prior to this, we were hoping it was installed on the system, and somehow that was ok. Now we're properly depending on it.

Lastly, I've patched `s2n-tls` so that it actually compiles. It's still not linking due to `-lgcc_eh`, due to static builds... but better something than nothing.

### DIFF
--- a/develop.nix
+++ b/develop.nix
@@ -79,6 +79,7 @@ includeos.pkgs.mkShell.override { inherit (includeos) stdenv; } rec {
         "${includeos.botan2.include}" \
         "${includeos.http-parser.include}" \
         "${includeos.lest.include}" \
+        "${includeos.libfdt.include}" \
         "${includeos.uzlib.include}" \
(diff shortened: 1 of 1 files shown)


mazunki:
There is a question about include-paths of dependencies. Do we strictly want all dependencies to be placed in a subdirectory (unless, of course, this is already done upstream)? This seems to be a bit inconsistent.

mazunki:
Rebased to fix merge conflicts, but there's still the question of whether we want to strictly enforce including headers coming from dependency as `#include <dependency/header.hpp>`. I have separated `include` and `include_root` for this purpose.

PR #2330, opened 2025-10-23
bump to c++23

mazunki:
It works! All tests are passing!

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -37,7 +37,7 @@ FILE(WRITE ${CMAKE_BINARY_DIR}/version.h
 
 #include_directories(${CMAKE_BINARY_DIR})
 
-set(CMAKE_C_FLAGS "-g -O0 -std=c11 -Wall -Wextra")
+set(CMAKE_C_FLAGS "-g -O0 -std=c23 -Wall -Wextra")
 
 set(NO_INFO "-DNO_INFO=1")
 if(INFO)


alfreb:
Love it!

mazunki:
```cpp
#include <new>
#include <cstdio>
#include <print>

struct Count { int n; };

int main() {
    const Count* counter = new const Count{3};
    int apples = counter->n;

    new (const_cast<Count*>(counter)) const Count{5};

    int bananas = counter->n; // UB
    std::println("apples={}", apples);
    std::println("bananas={}", bananas);
}
```

```shell
$ clang++ -Wall -Wextra -Wpedantic -O3 -std=c++23 -fsanitize=address,undefined main.cpp
$ ./a.out                                                                              
apples=3
bananas=5

=================================================================
==7668==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 4 byte(s) in 1 object(s) allocated from:
    #0 0x562bad2c87c1 in operator new(unsigned long) (/home/maz/nyaa/launder/a.out+0x15e7c1)
    #1 0x562bad2c9c6d in main (/home/maz/nyaa/launder/a.out+0x15fc6d)

SUMMARY: AddressSanitizer: 4 byte(s) leaked in 1 allocation(s).
```

mazunki:
> And the original warning I assume you got, which prompted you to introduce launder in the first place, would be very helpful

The warning which made me look into this is `aligned_storage_t` being deprecated. I'm trying to get a warning of the UB itself, but can't find a way to have the compiler tell me that other than through runtime sanitation.

Related: https://www.think-cell.com/assets/en/career/talks/pdf/think-cell_talk_lifetime.pdf

mazunki:
Cleanly rebased due to conflicts. Still good on tests.

alfreb:
> This example is only leaking sometimes for me, lol
> 
> ```c++
> #include <new>
> #include <cstdio>
> #include <print>
> 
> int main() {
>     const int* ptr = new const int(11); // provenance A
>     std::destroy_at(ptr);
> 
>     void* raw = const_cast<void*>(static_cast<const void*>(ptr));
> 
>     int* new_ptr = ::new (raw) int(42); // provenance B
> 
>     std::print("*new_ptr = {}\n", *new_ptr);       // okay
>     std::print("*ptr = {}\n", *ptr);               // UB
>     std::print("*ptr = {}\n", *std::launder(ptr)); // okay
> }
> ```
> 
> Interesting bit: if I use `println` instead of `print`, it always fails. Gotta love some UB.
> 
> `clang++ -Wall -Wextra -Wpedantic -O3 -std=c++23 -fsanitize=address,undefined -ggdb3 main.cpp`

Ok, so this makes a lot more sense - and it's completely different from the first example. Right? I think it's important that we understand anything new we introduce to the kernel properly. 

In the alloc / realloc example literally nothing changed, unless the pointer returned from realloc changed. And if the pointer changed, there's nothing std::launder could do for us. So the only examples we should consider are examples where the pointer itself doesn't change but what is stored at that pointer does. Agreed?

alfreb:
We had a good discussion offline. These are the two cases for `std::launder` according to cppreference (copy-pasting to preserve for the future): 
```C++
#include <cassert>
#include <cstddef>
#include <new>
 
struct Base
{
    virtual int transmogrify();
};
 
struct Derived : Base
{
    int transmogrify() override
    {
        new(this) Base;
        return 2;
    }
};
 
int Base::transmogrify()
{
    new(this) Derived;
    return 1;
}
 
static_assert(sizeof(Derived) == sizeof(Base));
 
int main()
{
    // Case 1: the new object failed to be transparently replaceable because
    // it is a base subobject but the old object is a complete object.
    Base base;
    int n = base.transmogrify();
    // int m = base.transmogrify(); // undefined behavior
    int m = std::launder(&base)->transmogrify(); // OK
    [assert](https://en.cppreference.com/w/cpp/error/assert.html)(m + n == 3);
 
    // Case 2: access to a new object whose storage is provided
    // by a byte array through a pointer to the array.
    struct Y { int z; };
    alignas(Y) [std::byte](https://en.cppreference.com/w/cpp/types/byte.html) s[sizeof(Y)];
    Y* q = new(&s) Y{2};
    const int f = reinterpret_cast<Y*>(&s)->z; // Class member access is undefined
                                               // behavior: reinterpret_cast<Y*>(&s)
                                               // has value "pointer to s" and does
                                               // not point to a Y object
    const int g = q->z; // OK
    const int h = std::launder(reinterpret_cast<Y*>(&s))->z; // OK
 
    [](...){}(f, g, h); // evokes [[maybe_unused]] effect
}
```

https://en.cppreference.com/w/cpp/utility/launder.html


I think it's absolutely terrible that `reinterpret_cast<Y*>(&s)` is not a pointer to Y*, making this UB;
```
  const int f = reinterpret_cast<Y*>(&s)->z; 
```
even if _we can guarantee_ that there is in fact a valid `Y` at the address pointed to by &s, because we just put it there with placement new. 

I also think it's scary that using `std::launder` when not every condition on the list of requirements is fulfilled, it can itself cause UB, but I guess that's a consequence of the fact that in the end it's just a pointer and the compiler can't have full knowledge of what a given pointer points to at any given point in time because of the halting problem. 

This is apparently OK: 
```
    const int h = std::launder(reinterpret_cast<Y*>(&s))->z; // OK
```

In addition to the clunky cast, we now also have to reassure the compiler that yes, I know what I'm doing, I'm both telling you that there is in fact a valid Y at this address, _and_ that the Y at that address is still alive. In two separate statements. Should I also sign a form claiming legal responsibility?

Anyway, launder it is. I don't fully understand it, but thanks @mazunki for finding this new protective gear against the dangers I wish we didn't have to know about.

PR #2338, opened 2025-10-28
Introduce `util::format`

mazunki:
As requested in https://github.com/includeos/IncludeOS/pull/2309#discussion_r2441846402, byte-formatters now reside in a standalone namespace.

I have also introduced a stringifier for `errno` values. The standard has `strerror()` for the corresponding message, but not the name. This provides `errno_name(errno)` for this purpose.

There's also a convenience helper for things that don't support string-views: `errno_cstr(errno)`. This is preferred over explicitly doing `errno_name(errno).data()` in corresponding locations because this way we can eventually deprecate c-strings from a single location without having to track down who uses printf-style specifiers.

Sadly, there is no convenient way to put these into an enum class and have the compiler turn it into a string. There are hopes for this in C++26, see: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2996r13.html

### DIFF
--- a/api/util/errno.hpp
+++ b/api/util/errno.hpp
@@ -1,0 +1,160 @@
+#ifndef UTIL_ERRNO_HPP
+#define UTIL_ERRNO_HPP
+
+#include <string_view>
+#include <errno.h>
+
+namespace util::format {
+
+  constexpr std::string_view errno_name(int e) noexcept {
+    switch (e) {  // list generated from musl's errno.h
+      case 0: return "OK";
+      case EPERM: return "EPERM";
+      case ENOENT: return "ENOENT";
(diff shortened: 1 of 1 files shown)


mazunki:
Actually, we have `api/utils/units.hpp`, which I had forgot someone mentioned. While `Byte_r` is alright, it doesn't provide `std::formatter<Byte_r>` which is desired for `std::format()`.

There are currently two different ways to represent bytes, this PR doesn't change that; none of which are friendly to formatting.

For what it's worth, we might want to introduce `Errno(n)` for the same purpose, so we can have `std::formatter<Errno>` too.

alfreb:
lgtm

PR #2339, opened 2025-10-28
Relocate memory-allocation specific files to their own path

mazunki:
As I will be adding more allocators going forward, it's useful to have a dedicated location for different allocators. This creates the `mem/` and `mem/alloc/` include paths for this purpose.

This also matches more closely to the namespace `os::mem` we were already using.

I suppose putting those files in `util/` was initially a temporary solution which just stuck around.


mazunki:
> Needs rebase, but lgtm

Done, tests are still 🌈

PR #2343, opened 2025-10-28
Segregate `kernel/memory.hpp` into meaningful parts

mazunki:
The old `kernel/memory.hpp` had several responsibilities, and albeit not being super long it felt hard to read. It comprised at least two fairly different parts:
- Setting up a mapping from virtual memory to physical pages (now `mem/vmap.hpp`)
- Setting up the allocator for the heap (now `mem/alloc.hpp`)

Furthermore, I've also created `mem/flags.hpp` since it doesn't quite fit either of them entirely, and it facilitates plans going forward.

Currently, we're assuming architecture-specific/hardware protection flags match IncludeOS-allocation flags. While this is currently the case, I do intend to change this in the future for the sake of type safety. `src/arch/x86_64/paging.cpp` uses `os::mem::Access` referring to hardware flags, while `src/kernel/multiboot.cpp` or `src/kernel/elf.cpp` refers to IncludeOS semantics. I intend to create a per-architecture translation mapping for this (`constexpr`, so it won't affect any performance, of course).

The responsibility separation between `mem/vmap` and `mem/alloc` seems to be represented in the usages across the code too (most of the results are non-overlapping):

```shell
$ rg mem/vmap.hpp | wc -l
18
$ rg mem/alloc.hpp | wc -l
14
```

This PR essentially makes #2329 obsolete.

### DIFF
--- a/api/mem/vmap.hpp
+++ b/api/mem/vmap.hpp
@@ -1,46 +1,17 @@
-// -*- C++ -*-
-// This file is a part of the IncludeOS unikernel - www.includeos.org
-//
-// Copyright 2017 Oslo and Akershus University College of Applied Sciences
-// and Alfred Bratterud
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-#ifndef MEM_MEMORY_HPP
-#define MEM_MEMORY_HPP
+#ifndef MEM_VMAP_HPP
+#define MEM_VMAP_HPP
 
 #include <util/bitops.hpp>
 #include <util/units.hpp>
-#include <mem/alloc/buddy.hpp>
-#include <mem/allocator.hpp>
 #include <mem/flags.hpp>
-#include <sstream>
-#include <expects>
 #include <mem/memmap.hpp>
+#include <expects>
+#include <algorithm>
+#include <cstdio>
+#include <stdexcept>
+#include <string>
 
 namespace os::mem {
-
-  using Raw_allocator = buddy::Alloc<false>;
-
-  /** Get default allocator for untyped allocations */
-  Raw_allocator& raw_allocator();
-
-  template <typename T>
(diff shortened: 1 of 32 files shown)


api/mem/vmap.hpp
```diff
@@ -253,93 +196,73 @@ namespace os::mem {
...
-    int len = snprintf(buffer, sizeof(buffer),
-            "%p -> %p, size %s, flags %#x",
-            (void*) lin,
-            (void*) phys,
-            util::Byte_r(size).to_string().c_str(),
-            (int) flags);
-
-    const bool isseq = __builtin_popcount(page_sizes) == 1;
+
+    std::string s = std::format( "lin {} -> phys {}, size {}, flags {:#x}",
```
alfreb: So using `std::string` means this will only work if the default allocator is ready at this point, because constructing a string allocates. Are we going to commit to that or is there a scenario where you'd want to do memory mapping before the default allocator is ready?

src/musl/mmap.cpp
```diff
@@ -1,124 +1,172 @@
...
-  // (address is in any case ignored)
+  if (util::missing_flag(flags, Sharing::Anonymous)) {
+    Expectsf(false, "Support for non-MAP_ANONYMOUS mappings is not yet implemented");
+    return mmap_failed(ENOTSUP);
+  }
 
-  auto* res = kalloc(length);
+  void *res = nullptr;
+
+  if (util::has_flag(flags, Flags::FixedOverride) or util::has_flag(flags, Flags::FixedFriendly)) {
```
alfreb: Wait - this is not just reorganizing stuff it it? Is this copy-pasted in from somewhere else or is it new functionality?

api/mem/alloc/arena.hpp
alfreb: Is this a new allocator? The PR says that this is just splitting a header into multiple parts. If you want to commit a new allocator that has to come in its own PR with unit tests and also be plugged in to an integration test.

api/util/errno.hpp
alfreb: This is reinventing `strerror` - please separate out if you think that's the right way to go and explain why.

alfreb (changes requested):
This does a lot more than it says on the label. Adds a new allocator without any tests, introduces duplicate string representations of error codes, introduces new mmap functionality in a C patch - lots of things. We can't bundle all this together.

api/util/errno.hpp
mazunki: Yep. I did not intend to add this here. The difference between this and `strerror` is tracked in #2338.

src/musl/mmap.cpp
mazunki: Yep, these changes will come later.

mazunki:
Welp, this should be much cleaner and easier to review now. I was continuing work on the same branch without branching out, and accidentally pushed some unfinished changes here without realizing.

PR #2347, opened 2025-10-28
Disable openssl

torgeiru:
Currently there are no tests for this module and it introduces warnings.


MagnusS (approved):
I think it's fine to disable this for now, as it's not working properly (tracked in #2348)

PR #2349, opened 2025-10-29
Formatting changes for dependencies

mazunki:
Requested change for #2325.

This purely introduces consistency changes and formatting.


MagnusS:
Thanks! Why is it better to use let..in here?

PR #2350, opened 2025-10-29
Removing arch_prctl from syscall and creating architecture specific musl calls

torgeiru:
Set_thread_area and arch_prctl are functions
used for setting the thread pointer. This is used
for thread local storage.

Currently arch_prctl goes through a system call trap. This patches musl so that it goes through a typical function call instead.

Also, this moves the implementation of these
function calls to architecture, removing the need for macros.


api/musl/common.hpp
alfreb: I'm not sure we want this to expose this. In the API folder you want to have anything you are willing to support long term, and that you actively think is a good idea for service developers to use. That said - if you can provide a good reason for why someone implementing a service would want to use strace to wrap their own stuff, maybe this should be renamed `strace.hpp` instead. But in general I think our API should get smaller, not bigger.

src/arch/aarch64/arch_musl.cpp
```diff
@@ -0,0 +1,27 @@
...
+}
+
+extern "C"
+long syscall_SYS_set_thread_area(struct user_desc *u_desc) {
+    return strace(sys_set_thread_area, "set_thread_area", u_desc);
+}
+
+extern "C"
+long sys_arch_prctl(int code, uintptr_t ptr) {
+    os::panic("This should not happen!");
```
alfreb: Please explain why. Like because multithreading is not supported on this arch or something. As helpful as you can.

src/arch/x86_64/syscall_entry.cpp
```diff
@@ -14,63 +14,15 @@
...
 uintptr_t syscall_entry(uint64_t n, uint64_t a1, uint64_t a2, uint64_t a3,
                    uint64_t a4, uint64_t a5)
 {
-  switch(n) {
-  case 158: // arch_prctl
-    sys_prctl(a1, a2);
-    break;
-  default:
-    kprintf("<syscall entry> no %lu (a1=%#lx a2=%#lx a3=%#lx a4=%#lx a5=%#lx) \n",
+  kprintf("<syscall entry> no %lu (a1=%#lx a2=%#lx a3=%#lx a4=%#lx a5=%#lx) \n",
```
alfreb: Should we panic here? I don't think we ever want to use the syscall instruction so if some library does (like if libc decides to introduce it in more places) this will loop pretty mysterious. Not a dealbreaker though - this is what the kernel is doing today, so it's fine. But if you agree maybe use the opportunity to fail early.

PR #2354, opened 2025-10-30
Replace libfdt's CMakeLists.txt with nix build

mazunki:
I'm splitting up #2325 into standalone PRs for easier reviewing.

This was the only dependency which wasn't using `default.nix` in our list. It should be easier to manage our targets when everything follows the same structure.


elstr-512:
> @elstr-512 , does this help you?

Probably! My current solution isn't great, I'm just inserting the libfdt/dtc dependency in  [`overlay.nix`](https://github.com/elstr-512/IncludeOS/blob/066352b874ebd2cb076b008366d8b3bf0cc6cc80/overlay.nix).

``` nix
   ...
      aarch64_inputs =
          if self.stdenv.targetPlatform.system == "aarch64-linux" then [
            prev.pkgsStatic.dtc
          ]
          else [];
   ...
```

MagnusS:
From @elstr-512's branch it looks like we can use dtc from nixpkgs here so we don't have to maintain our own nix build for libfdt. I suggest we close this PR for now, then we can revisit when @elstr-512 submits the ARM build patches. It looks like ./deps/libfdt could then be removed completely.

PR #2360, opened 2026-03-24
Add flake support

mazunki:
Depends on https://github.com/includeos/vmrunner/pull/49.

This PR adds flake support to IncludeOS, allowing users to build the kernel, and build+run their unikernel through the flake interface. This reduces the pipeline of running the example unikernel to `nix run .#example`, which handles all the rebuilding.

Furthermore, this allows standalone unikernels to be able to make their own `nix run` wrappers which will propagate inputs automatically. 

There are no changes for non-flake users, and tests are still using the old mechanisms. This support is just added as a convenience, for now.  We expose the following:
- `nix build` which simply rebuilds IncludeOS
- `nix build .#chainloader` will build the chainloader
- `nix build .#example` will build the unikernel from `./example/`.
- `nix run .#example` builds and runs the example unikernel image.
- `nix run . -- ./path/to/result` will run `boot -j vm.json result.elf.bin` from an already built unikernel.

This approach allows us to override inputs both locally and remotely:
- `--override-input vmrunner ~/path/to/vmrunner`
- `--override-input vmrunner github:mazunki/vmrunner`

And IncludeOS itself doesn't need to be cloned. This "just works": `nix run github:mazunki/includeos/flakify#example`.

Note that `mkUnikernel.nix` is largely the same as `unikernel.nix`, but rewritten in a way that makes sense for flakes. Eventually, I'm not sure it makes sense to keep both.


mazunki:
For reference, this is a flake.nix I use in a unikernel:
```
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/25.05";
    includeos.url = "github:mazunki/IncludeOS/flakify";
    # includeos.url = "github:includeos/IncludeOS";
    vmrunner.url = "github:mazunki/vmrunner/flakify";
    # vmrunner.url = "github:includeos/vmrunner";

    includeos.inputs.nixpkgs.follows = "nixpkgs";
    includeos.inputs.vmrunner.follows = "vmrunner";
    vmrunner.inputs.nixpkgs.follows = "nixpkgs";
  };
  outputs = { self, nixpkgs, includeos, vmrunner }:
    let
      system = "x86_64-linux";
      mkUnikernel = includeos.lib.${system}.mkUnikernel;
    in {
      packages.${system}.default = mkUnikernel {
        unikernel = ./src;
        vmrunner = vmrunner.packages.${system}.default;
      };

      apps.${system}.default = {
        type = "app";
        program = "${nixpkgs.legacyPackages.${system}.writeShellScript "run" ''
          set -e
          built=$(nix build --no-link --print-out-paths)
          exec ${includeos.apps.${system}.boot-unikernel.program} "$built" "$@"
        ''}";
      };
    };
}
```

PR #2361, opened 2026-03-27
support booting with more than 4G of memory

mazunki:
This PR adds support for booting IncludeOS with more than 4 GB of physical memory, updating the example unikernel's message to show these mappings. Before this PR, the max size of the heap would be ~2.2GiB.

The old example unikernel wasn't terribly useful: maybe we could extend the example unikernel to be informative of more aspects of the kernel's state after it has been booted?

It additionally adds preemptive supports to unassigning virtual mappings, and shuts down the unikernel if a strace condition has been hit.

See full commit messages for specific details.


mazunki:
Note that this solution isn't perfect: LiveUpdate and other components should be made dynamic and cooperate somehow with selecting the heap's size. The preemptive support of unassigning memory map ranges aims to facilitate this transition.

mazunki:
Integration tests are currently unable to run `kernel/kprint` and `net/http` [on account of some missing python modules](https://github.com/includeos/IncludeOS/issues/2362). The rest of tests are ok.

Pending introducing integration tests for differently sized `vm.json`.

example/src/vm.json
```diff
@@ -0,0 +1,3 @@
+{
+  "mem": 16384
```
MagnusS: This is a lot - could we make it smaller?

src/arch/x86_64/paging.cpp
```diff
@@ -376,6 +376,8 @@ uintptr_t mem::active_page_size(uintptr_t addr){
 
 void allow_executable()
 {
+  // FIXME: this seems weird. is this only used for tests?
```
MagnusS: This is used for the executable part of the IncludeOS service binary

src/arch/x86_64/paging.cpp
```diff
@@ -391,7 +393,7 @@ void allow_executable()
   m.page_sizes = os::mem::Map::any_size;
   m.flags      = os::mem::Access::execute | os::mem::Access::read;
 
-  os::mem::map(m, "ELF .text");
+  os::mem::map(m, "Executable");
```
MagnusS: Perhaps the text could be "ELF executable"

src/musl/common.hpp
```diff
@@ -74,6 +74,7 @@ template<typename Fn, typename ...Args>
 inline auto strace(Fn func, [[maybe_unused]]const char* name, Args&&... args) {
   if (!kernel::state().allow_syscalls) {
     fprintf(stderr, "Syscalls not allowed here. Unexpected call to %s - terminating\n", name);
+    os::halt();
```
MagnusS: Instead of halt here it would better to fix the Expect-call so we can get stack traces when this happens after boot (and heap is available)

src/platform/x86_pc/os.cpp
```diff
@@ -145,13 +143,28 @@ void kernel::start(uint32_t boot_magic, uint32_t boot_addr)
...
+  {
+    if (not entry.is_available()) continue;
+
+    if (entry.len > heap_map.len) {
+      heap_map = entry;
+    }
+  }
+  uintptr_t end = heap_map.addr + heap_map.len - 1;
+
+  if (heap_map.addr < 0x1'000'000) {
```
MagnusS, line 159: Could this be solved by adding the LiveUpdate and Syslog ranges to the map statically? Should at least be commented here why there's a check

MagnusS (changes requested):
Looks good, but some smaller comments inline :-) Thanks!

src/musl/common.hpp
mazunki: This change has been moved to another PR.

mazunki:
Integration tests have been added. This should be ready to be merged now (after #2363).

Some of the tests are in a half-disabled state currently because the upstream buddy allocator doesn't support large allocations. The integration itself requires 6 GiB on the system currently: anything less wouldn't really make sense for this test. 16 GiB tests are desired cuz it seems like buddy breaks apart after around a 8GiB limit.

Note:
- f0cacb3 fixes up 02cbf1561.
- This also builds on #2363  cuz of older tests.

mazunki:
attempting to autosquash 9b641df in a rebase causes a bunch of commits here, so i've left it as a standalone commit.

MagnusS:
Thanks! I ran the tests locally and all looks good

PR #2363, opened 2026-03-27
remove deprecated python2 imports

### LINKED ISSUE #2362: Python modules not found in some tests
Tests 5.4 and 8.4 are unable to run due to missing python modules:

`Step 5.4 failed (nix-build ./unikernel.nix  --argstr unikernel ./test/integration/kernel/kprint --arg doCheck true)`:

```
File "/nix/store/a75ma10zxr7z9ym9q5hcqphcq39s6d66-source/test.py", line 6, in <module>
    from past.utils import old_div
ModuleNotFoundError: No module named 'past'
```
`👷⛔ Step 8.4 failed (nix-shell ./unikernel.nix  --argstr unikernel ./test/integration/net/http --arg doCheck true)`:
```
  File "/home/maz/repos/includeos/IncludeOS/test/integration/net/http/test.py", line 3, in <module>
    from future import standard_library
ModuleNotFoundError: No module named 'future'
```

I'm not sure exactly when this started happening. I don't suppose it happened cuz of https://github.com/includeos/vmrunner/pull/47... but rather it would be because of some changes in nixpkgs...? I thought we were pinning nixpkgs to avoid these issues, though.

mazunki:
Some tests were using imports that permit compatibility with python2. They are no longer necessary, and causes tests to fail.

All tests pass with these changes.

Closes #2362.

### DIFF
--- a/test/integration/kernel/kprint/test.py
+++ b/test/integration/kernel/kprint/test.py
@@ -3,7 +3,6 @@
 from __future__ import division
 from __future__ import print_function
 from builtins import str
-from past.utils import old_div
 import sys
 import os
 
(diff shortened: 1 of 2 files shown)


alfreb:
Thanks!

Issue #2364, opened 2026-04-08
`os::shutdown()` and `os::halt()` are misleading


mazunki:
Calling `os::shutdown()` doesn't shut down the machine when run with vmrunner because of a missing end-of-transmission (`\x04`) signal over the wire. To actually shut down the machine, we currently have to call `__arch_poweroff()`.

`os::halt()` is a confusing name, despite matching the `hlt` instruction of x86. I feel this should be renamed to something like `os::yield()`,  `os::pause()`, or `os::wait_for_interrupt()`. As it stands, without reading the implementation, it's easy to assume it's equivalent to shutting down the machine.

While it would be easy to think `os::shutdown()` should call `__arch_poweroff()`, instead of only setting the state of the kernel as non-running, it seems like the semantic meaning of "shutting down" means that there are no more events pending on the unikernel (see `src/platform/x86_pc/os.cpp`). Is this correct? Feels confusing.

Note that `__arch_poweroff()` and `__arch_reboot()` are actually exposed through `api/arch.hpp`, so services can call this if they need to. That feels like an implementation detail that shouldn't be exposed like that, but rather through `os::shutdown()` (or perhaps `os::poweroff()`, if shutdown's name means something else).

Furthermore, it's now expected that services shut down the os themselves. There is a weak `Service::stop()` function with an empty implementation that is called after the kernel stops running. My first intuition was that this would work, but as explained, this function is actually called after the service has stopped running, which is not equivalent to the service having returned:
```cpp
#include <os>

void Service::start(){ std::println("Hello :)"); }
void Service::stop(){ os::shutdown(); }
```

We advertise that this should work, which leaves the machine hanging:
```cpp
#include <os>

int main(){
  std::println("Hello :)");
}
```

mazunki:
`api/hal/machine.hpp` declares `os::Machine::poweroff`, which doesn't seem to be defined anywhere. I believe most `__arch_poweroff()` calls should probably be replaced by `os::machine()::poweroff()` as a wrapper?

PR #2365, opened 2026-04-08
automatically shut down service after main function returns

mazunki:
Builds and tests succesfully on top of #2363 due to tests.

This PR addresses a point in #2364 where the README advertises that the following program is a minimal valid program:

```cpp
#include <os>
int main() {
  std::println("Hello");
}
```

Before, this program would leave the kernel hanging, without shutting down.

If `Service::start()` is defined, on the other hand, this responsibility remains on the service.


mazunki:
rebased to sync with main

mazunki:
Re-rebased and tested. :)

PR #2366, opened 2026-04-08
prevent cyclic Expects warning

mazunki:
This builds on top of #2363 due to tests.

As requested in https://github.com/includeos/IncludeOS/pull/2361#discussion_r3000771766, it's better to handle cyclic faults at the exception handler. This PR fixes this, with some practical compromises.

Ideally, I would have liked to call `os::panic()`, but this has some issues: currently, `os::panic` only supports being called with `const char *why`, with no concept of source location or expressions.

Furthermore, it (used to) assume libc was initialized (i.e. calling `printf` and friends). I've added a check to prevent this being an issue, irrespective of the problem this PR solves. I think `os::panic` should have two interfaces: a dumb one that can be used irrespective of initialization state (that is, a simple fallback mechanism that is similar to my implementation of `__expect_fail`), and a complete one that accepts `std::format_string` and cooperates correctly with the rest of the system.

Unrelated to this, calling `__arch_poweroff()` does not feel good. I wanted to call `os::shutdown()`, but this doesn't actually trigger a shutdown unless the proper message has been sent to vmrunner. I've opened #2364 with more details about this.

### DIFF
--- a/api/arch.hpp
+++ b/api/arch.hpp
@@ -27,8 +27,8 @@
 #include <string>
 
 
-extern void __arch_poweroff();
-extern void __arch_reboot();
+[[noreturn]] extern void __arch_poweroff();
+[[noreturn]] extern void __arch_reboot();
 extern void __arch_enable_legacy_irq(uint8_t);
 extern void __arch_disable_legacy_irq(uint8_t);
 extern void __arch_system_deactivate();
--- a/api/expects
+++ b/api/expects
@@ -32,6 +32,19 @@
 #endif
 
 #include <os.hpp>
+#include <serial>
+
+inline bool __expect_is_recursive() {
+#ifndef UNITTESTS
+  static volatile bool in_failure = false;
+  if (in_failure) return true;
+  in_failure = true;
+  return false;
+#else
+  return false;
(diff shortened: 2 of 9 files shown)


mazunki:
rebased to sync with main

alfreb:
This looks good - but can you help me figure out how to test it? I will, but might not be today. Also, I agree that we should have an early and late panic, where only the late one expects libc to be initialized- although we could just check if it is. I think there's a boolean for that somewhere.

mazunki:
We do have `kernel::libc_initialized()`, but I'm not sure that accounts for all early initialization being ready: [93efffa](https://github.com/includeos/IncludeOS/pull/2366/commits/93efffa26d9a1db94b1a8a2683680c03f1350b8c) does exactly that.

I tested this by manually lying in `src/kernel/multiboot.cpp::_multiboot_memory_end()` (i.e. immediately `return os::Arch::max_canonical_addr;`)  . I'm not sure if there's a clean way to test this, as this should really only trigger if something is /really wrong/ with the early boot process.

mazunki:
re-based and re-tested.

Issue #2333, opened 2025-10-27
Use enum-class for PCI configuration registers


MagnusS:
They are currently defines in https://github.com/includeos/IncludeOS/blob/main/api/hw/pci_device.hpp#L25, but it would be better to move them into an enum class for type safety.

See discussion in https://github.com/includeos/IncludeOS/pull/2328/files#r2448877338

Issue #1409, opened 2017-07-26
VirtioNet mac address it not randomized


fwsGonzo:
This creates problems for people who want to run many instances locally. Creating multiple vm.jsons for this is unnecessary as the address should be random by default.

Issue #2352, opened 2025-10-29
De-x86-ify IncludeOS for upcoming ARM support


torgeiru:
Lots of the code assumes x86 for example PCI driver code (port IO)
and there are many `asm volatile`. I think we should switch from these to intrinsics and C++-standard.

I must probably help with PCI driver code.

Here is a comprehensive references of assembly instructions:
```log
api/smp_utils:    asm("mfence");
api/smp_utils:      asm("pause; nop;");
api/debug:      asm("mov $0,%eax ; 0: ; test %eax,%eax ; jz 0b"); }
api/smp:    asm("movl %%gs:(0x0), %0" : "=r" (cpuid));
api/smp:    asm("movl %%fs:(0x0), %0" : "=r" (cpuid));
api/arch/x86_64.hpp:  asm("rdtsc" : "=a"(lo), "=d"(hi));
api/arch/i686.hpp:  asm("rdtsc" : "=A" (ret));
lib/LiveUpdate/src/hotswap.cpp:asm(".org 0x8000");
test/integration/mod/gsl/service.cpp:          THEN("span saves us from complete embarrasment") {
test/integration/stl/coroutines/service.cpp:  asm("pause");
test/integration/kernel/fiber/fiber_smp.cpp:      asm("pause");
test/integration/kernel/fiber/fiber_smp.cpp:        asm("pause");
test/integration/kernel/fiber/fiber_smp.cpp:      asm("pause");
test/integration/kernel/fiber/fiber_smp.cpp:        asm("hlt");
test/integration/kernel/fiber/fiber_smp.cpp:    asm("pause");
test/integration/kernel/fiber/fiber_smp.cpp:    asm("pause");
test/integration/kernel/fiber/fiber_smp.cpp:  asm("pause");
test/integration/kernel/exception/service.cpp:  //asm ("movl $0, %eax");
test/integration/kernel/exception/service.cpp:  //asm ("idivl %eax");
test/integration/kernel/exception/service.cpp:  asm ("movq $0, %rax");
test/integration/kernel/exception/service.cpp:  asm ("idivq %rax");
src/platform/x86_nano/platform.cpp:  while (1) asm("cli; hlt;");
src/platform/x86_nano/platform.cpp:  asm("hlt");
src/platform/x86_pc/init_libc.hpp:#define LL_ASSERT(X) if (!(X)) { kprint("Early assertion failed: " #X "\n");  asm("cli;hlt"); }
[...]

PR #2367, opened 2026-06-10
typed PCI registers, virtio MAC randomization, SMP syscall trap

uadhran:
**Tested:** `nix-build unittests.nix` — 85/85 unit tests passed (Nix 2.34.7, Linux).

Not run yet: `./test/test.sh`, integration/unikernel builds.

---

Three small fixes:

**PCI config registers (#2333)**
PCI register offsets were plain #defines in the header. Moved them to
enum classes (`config_reg`, `command`, `cap_id`) for type safety.
Updates in `pci_device.cpp` and `pci_msi.cpp`.

**VirtioNet MAC (#1409)**
QEMU hands out the same default MAC on every virtio-net device, which is
awkward when running multiple instances locally. After reading the MAC
from virtio config we randomize the host-specific bytes and set the
locally-administered bit. Uses RDRAND/RDSEED when available, with a
pci_addr + clock fallback (drivers init before `RNG::init()`).

**SMP syscall trap (#2358)**
Only CPU 0 had the syscall MSR set up. Pulled that into
`init_syscall_trap()` and call it from both the BSP libc init path
and `revenant_main`. Handler still panics — syscalls aren't
implemented — but APs get a trap instead of nothing.

Related: #2333, #1409, #2358

### DIFF
--- a/api/hw/pci_device.hpp
+++ b/api/hw/pci_device.hpp
@@ -22,33 +22,46 @@
 #include <vector>
 #include <unordered_map>
 
-/* PCI Register Config Space */
-#define PCI_DEV_VEND_REG 0x00	/* for the 32 bit read of dev/vend */
-#define PCI_VENDID_REG   0x00
-#define PCI_DEVID_REG    0x02
-#define PCI_CMD_REG      0x04
-#define PCI_STATUS_REG   0x06
-#define PCI_REVID_REG    0x08
-#define PCI_PROGIF_REG   0x09
-#define PCI_SUBCLASS_REG 0x0a
-#define PCI_CLASS_REG    0x0b
-#define PCI_CLSZ_REG     0x0c
-#define PCI_LATTIM_REG   0x0d
-#define PCI_HEADER_REG   0x0e
-#define PCI_BIST_REG     0x0f
-#define PCI_CAPABILITY_REG 0x34
-
-#define PCI_COMMAND_IO		 0x01
-#define PCI_COMMAND_MEM		 0x02
-#define PCI_COMMAND_MASTER 0x04
-
-#define PCI_CAP_ID_VNDR      0x09
-#define PCI_CAP_ID_AF        0x13	/* PCI Advanced Features */
-#define PCI_CAP_ID_MAX       PCI_CAP_ID_AF
-#define PCI_EXT_CAP_ID_PASID 0x1B	/* Process Address Space ID */
-#define PCI_EXT_CAP_ID_MAX   PCI_EXT_CAP_ID_PASID
-
 namespace PCI {
+
+  /** PCI configuration space register offsets */
+  enum class config_reg : uint8_t {
+    DEV_VEND     = 0x00, /**< 32-bit device/vendor ID */
+    DEVID        = 0x02,
+    CMD          = 0x04,
+    STATUS       = 0x06,
+    REVID        = 0x08,
+    PROGIF       = 0x09,
+    SUBCLASS     = 0x0a,
+    CLASS        = 0x0b,
+    CLSZ         = 0x0c,
+    LATTIM       = 0x0d,
+    HEADER       = 0x0e,
+    BIST         = 0x0f,
+    CAPABILITY   = 0x34
+  };
+
+  /** PCI command register flags */
+  enum class command : uint16_t {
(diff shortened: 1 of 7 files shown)


torgeiru:
When looking statically PCI/enum and static casting seems fine. Syscall handler for SMP is very much wanted, but maybe the init function should be part of `src/arch`? The MAC stuff I am more unsure of but might be right. Source for the MAC/ring ideas? Unix tests pass for my run. Must run integration tests as well (darn networking tests).

uadhran:
Thanks for the review @torgeiru .

**PCI enums** — noted, glad the static casts look fine.

**SMP syscall trap** — agree that `init_syscall_trap()` probably belongs under `src/arch/x86_64/` next to `__syscall_entry` / `syscall_entry.cpp` rather than in `platform/x86_pc`. Happy to move it in a follow-up commit if you prefer that before merge.

**Virtio MAC** — addresses #1409: QEMU gives the same default virtio-net MAC for every instance, which breaks running multiple copies locally. After reading the MAC from virtio config we keep the prefix from the device and randomize the host-specific bytes (`addr[3..5]`), with the locally-administered bit set (`addr[0] |= 0x02`). Entropy uses RDSEED/RDRAND when available (same instructions IncludeOS already uses elsewhere); the fallback is `pci_addr` + time/cycles because drivers init before `RNG::init()`.

**Tests** — unit tests pass here as well (`nix-build unittests.nix`, 85/85). Working on integration tests next; will update the PR once those are done.

torgeiru:
I suppose MAC identifier generation is not crypto. But why did you make a distinct function for MAC RNG within VirtioNET? I believe this generation can be decoupled from the driver and reused by other drivers e.g. `E1000` and `VMXNET`. Also, shouldn't this function be decoupled from arch-specific entropy generation (I think this is the point of extract RNG function?).

As a side note: These commits/PRs can be broken into multiple distinct PRs. SMP and PCI are probably an instant accept but because it is coupled with MAC the merging is stalled (or would be if the maintainers weren't on vacation hehe). 

Edit:
See now that drivers are initialized before `RNG::init()` which is kinda messed up. 

I think these things could be fixed for follow up PRs. See #2352 for vision :)

mazunki:
Thanks for this! Nice cleanup. I just ran all the tests, and everything seems to be working fine. ✨ 

A few notes, though. Looking at the changes, it seems there are two separate PRs: one regarding the PCI devices, and one regarding the initialization of the syscall traps. Would it be inconvenient to separate this into two PRs? I know the maintainers have a preference for smaller PRs: they are quicker to review.

For the PCI devices, I don't think we have any proper integration tests (what was failing before this PR? maybe this is an opportunity to test for that?). We have `fs/virtio_block`, which checks parts of it indirectly.

Regarding syscall trap initialization, it always felt weird to me that this is part of the libc initialization. I suspect this is necessary cuz of the C++ runtime, but I haven't dug too deep into that. Nice to separate that, thanks.

Otherwise, I love the code style :) lgtm

uadhran:
Thanks @torgeiru @mazunki  — this is really helpful.
I'll split this up as suggested:
1. PCI enum-class (#2333) — on its own
2. SMP syscall trap init (#2358) — on its own, with init_syscall_trap() moved toward src/arch/x86_64/ next to the syscall entry code rather than living in platform/x86_pc

I'll pull the Virtio MAC work (#1409) out of this stack for a follow-up. I agree it shouldn't be a virtio-only helper with direct RDRAND/RDSEED calls — a shared MAC generation path that other drivers (E1000, VMXNET, etc.) can reuse makes more sense, and the RNG side should go through the proper extract path once init ordering is sorted (#2352). The driver-before-RNG::init() ordering is a real constraint; happy to tackle that properly in a separate change rather than papering over it in one driver.

@mazunki good to hear the full test suite passed on your side. I'll keep PRs small per project preference.

I'll update here once the split PRs are up and link them from this thread. We can close or repurpose this one once those are open.

uadhran:
Split done as discussed:

- #2369 — PCI enum-class (#2333)
- #2370 — SMP syscall trap on APs (#2358)

Virtio MAC (#1409) stays out of this stack for a follow-up PR with a shared driver helper / proper RNG path.

This PR can be closed in favour of the two above.

PR #2370, opened 2026-07-07
platform: set up syscall trap on SMP application processors

uadhran:
Only CPU 0 had the syscall MSRs configured. This pulls that setup into `x86::init_syscall_trap()` and calls it from both the BSP libc init path and `revenant_main`, so application processors get a trap instead of nothing. The handler still panics — syscalls aren't implemented.

**Changes:**
- `src/arch/x86_64/syscall_trap.cpp` — programs `IA32_STAR` and `IA32_LSTAR`
- `api/arch/x86/syscall.hpp` — declaration
- `src/platform/x86_pc/init_libc.cpp` — BSP calls `init_syscall_trap()`
- `src/platform/x86_pc/apic_revenant.cpp` — each AP calls `init_syscall_trap()`

**Tested:** `nix-build unittests.nix` — 85/85 passed (Nix 2.34.7, Linux).

Split from #2367 per review feedback.

Closes: #2358


uadhran:
Moved `init_syscall_trap()` into `src/arch/x86_64/syscall_trap.cpp` with declaration in `api/arch/x86/syscall.hpp`, per review. Platform code only calls it now.

api/arch/x86/syscall.hpp
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+namespace x86 {
```
mazunki, line 3: Do we want this under `x86` only? I think syscalls can exist on any architecture, and thus it would make sense to declare this for all architectures, and require architectures for which it makes no sense to initialize syscalls to stub an empty implementation.

Maybe @elstr-512 has an opinion on this?

src/platform/x86_pc/init_libc.cpp
```diff
@@ -153,18 +154,8 @@ namespace x86
...
-    uint64_t star_user_cs   = 8ull << 48;
-    uint64_t star = star_kernel_cs | star_user_cs;
-    x86::CPU::write_msr(IA32_STAR, star);
-    x86::CPU::write_msr(IA32_LSTAR, (uintptr_t)&__syscall_entry);
-  #elif defined(__i386__)
-    KDEBUG("Initialize syscall intr (32-bit)\n");
-    #warning Classical syscall interface missing for 32-bit
-  #endif
+    KDEBUG("* Initialize syscall trap\n");
+    x86::init_syscall_trap();
```
mazunki, line 158: I'm still wondering if this should be part of libc initialization. I wouldn't do such a change in this PR fwiw, but I would consider opening an issue to discuss it further, maybe.

mazunki:
This is much cleaner than what it used to be, thanks! All integration tests are passing too.

I kinda wish we could add some integration tests to verify syscalls fail to trap before they're initialized, and make sure they're trapped after initialization.

Otherwise, see inlined comments.

mazunki:
Btw, would merging this resolve #2358 (cc @torgeiru)? If so, you can add `Closes: #2358`  to the PR :)

uadhran:
Good point — a test that exercises syscall trapping before vs after `init_syscall_trap()` would make this much easier to regress-check.

I'd rather not fold that into this PR since it's already scoped to AP MSR setup. I'll open a follow-up issue for an integration test (or small test service) that verifies:
1. before init: syscall does not trap as expected on an AP, and
2. after init: syscall traps and hits the panic handler.

Happy to take that on once this lands, unless you'd prefer it block merge.

uadhran:
Yes — this should resolve #2358. I've added `Closes: #2358` to the PR description. Thanks for the nudge (and cc @torgeiru).

api/arch/x86/syscall.hpp
uadhran: Fair question. I kept this under `x86::` for this PR because the implementation is x86_64-specific (`IA32_STAR` / `IA32_LSTAR`) and IncludeOS only runs the syscall path on x86 PC today.

A broader shape would be something like `os::init_syscall_trap()` in a neutral header with arch stubs (no-op on aarch64, real impl on x86_64). That's a nice cleanup but a separate refactor from "APs also get the MSRs."

I've left it as-is here to keep the diff small. Happy to follow up with an arch-neutral API if that's the direction you and @elstr-512 prefer — would you want that as a follow-up PR after this merges?

src/platform/x86_pc/init_libc.cpp
uadhran: Agree — hooking syscall MSR setup off the libc init path is a bit odd, and I wouldn't want to reshuffle boot ordering inside this PR.

I'll open an issue to discuss when/where syscall trap init should live (e.g. earlier platform bring-up vs libc/runtime init, and how that interacts with AP startup). This PR only extracts the MSR programming and ensures BSP + APs both call the same helper; it doesn't change *when* the BSP first calls it.

Thanks for calling that out.

torgeiru:
Veri clean.

